PR #5, opened 2016-11-24
Rewrite to the new, private API

mxstbr:
With the new private API, all the IIFEs are unnecessary since we can just pass an object to the styled() calls. This commit updates the plugin and the tests to reflect the new result. Need to wait for styled-components/styled-components#227 (which introduces the new API) to be merged before releasing.

----

Please review the code @vdanchenkov, I'm not very familiar with writing babel plugins, so if you see things that could be done easier please let me know!

@geelen please review the result of the conversion in the `after.js` files of the fixtures!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,9 +4,9 @@ import hash from './utils/hash'
 import getTarget from './utils/get-target'
 
 const buildStyledCall = template(`styled({
-	target: TARGET,
-	displayName: DISPLAYNAME,
-	identifier: IDENTIFIER
+  target: TARGET,
+  displayName: DISPLAYNAME,
+  identifier: IDENTIFIER
 })`)
 
 const isStyled = (tag) => (tag.object && tag.object.name == 'styled') || (tag.callee && tag.callee.name == 'styled')
@@ -19,7 +19,7 @@ export default function({ types: t }) {
       TaggedTemplateExpression: {
         enter(path, { opts }) {
           const addDisplayName = (opts.displayName === undefined || opts.displayName === null) ? true : opts.displayName
-					const addIdentifier = (opts.ssr === undefined || opts.ssr === null) ? true : opts.ssr
+          const addIdentifier = (opts.ssr === undefined || opts.ssr === null) ? true : opts.ssr
           const tag = path.node.tag
 
           if (!isStyled(tag)) return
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -17,76 +19,74 @@ export default function({ types: t }) {
...
 
-            // identifiers are the only thing we can reliably get a name from
-            if (!t.isIdentifier(displayName)) {
-              displayName = undefined
-            } else {
-              displayName = displayName.name
-            }
+          // styled call without variable
+          if (!target) {
+            target = getTarget(path.node.tag)
```
vdanchenkov: getting target from path.node.tag works for all tests. In which cases you need to determine it differently?

src/index.js
```diff
@@ -17,76 +19,74 @@ export default function({ types: t }) {
...
+          })
+          // Put together the styled call with the template literal
+          // to get the finished styled({ })`` form! 🎉
+          const styledCall = t.taggedTemplateExpression(
+            call.expression,
+            path.node.quasi
+          )
+          path._styledComponentsSeen = true
+
+          path.replaceWith(styledCall)
```
vdanchenkov: As we got rid of wrapping, code above can be reduced to 
```
          // Put together the styled call with the template literal
          // to get the finished styled({ })`` form! 🎉
          path.node.tag = call.expression
```

No need to mark with _styledComponentsSeen

src/index.js
vdanchenkov: Yes, in my understanding it's always path.node.tag of TaggedTemplateExpression

vdanchenkov:
I've been thinking of similar change, but I thought of less intrusive conversion of `styled.div` to `styled('div', {displayName, identifier})` It would make plugin backward compatible with older styled components versions meaning it would not break them. Not sure if it's important.

mxstbr:
I don't think that's a concern, since we can just publish a new major version with a peerDep to the new styled-components version that supports it. I don't remember why we decided to not do it that way though, but now it doesn't really matter and I don't want to rewrite it 😉

src/index.js
mxstbr: You are right, thanks! Updated!

src/index.js
mxstbr: Amazing, thanks!

PR #6, opened 2016-11-26
Support different import names

### LINKED ISSUE #1: Support different import names
Currently, this breaks the babel plugin:

```JS
import s from 'styled-components'
```

Instead of hard-coding and checking for the `styled` keyword, we should be looking for the import declaration and getting the variable name. (what I do [here](https://github.com/styled-components/stylelint-processor-styled-components/blob/9d904c963466e397de2d08f84314249d9c0e7ce9/src/index.js#L60-L63) and [here](https://github.com/styled-components/stylelint-processor-styled-components/blob/9d904c963466e397de2d08f84314249d9c0e7ce9/src/utils/parse.js#L5-L23) in the stylelint processor)

mxstbr:
Closes #1 

Please review @vdanchenkov!


vdanchenkov:
Variable name can be passed to template, to build it only once.
```jsx
const buildStyledCall = template(`STYLED({
  target: TARGET,
  displayName: DISPLAYNAME,
  identifier: IDENTIFIER
})`)

const call = buildStyledCall({
  STYLED: t.identifier(importedVariableName),
  TARGET: target,
  DISPLAYNAME: (addDisplayName && t.stringLiteral(displayName)) || t.identifier('undefined'),
  IDENTIFIER: (addIdentifier && t.stringLiteral(identifier)) || t.identifier('undefined')
})
```

mxstbr:
Good call, fixed!

PR #8, opened 2016-11-28
use filename for displayName

vdanchenkov:
@mxstbr something like this?


mxstbr:
We might need to uppercase it? `before` => `Before`?

vdanchenkov:
I think, I'd prefer to keep original case of file name to not surprise user. But in my projects I uppercase component files, so it doesn't really matter for me. If you think it's better to uppercase I'll add it.

mxstbr:
No you're actually right, good point! Should we maybe add the extension as well to make it clear that this is the filename? `before.js__Component-asdf123` is clearer to me where the `before` comes from than just `before__Component-asdf123`?

mxstbr:
Also, should we make the options consistent? These two different capitalizations are kind of weird:

```JS
["styled-components", {
  filename: false,
  displayName: false,
}]
```

We should choose one convention and go with that?

vdanchenkov:
It seems reasonable when component does have it's own name. But let's get back to use case I started from.

```jsx
// in FancyButton.js 
import styled from 'styled-components'
export default styled.button` `
```

I feel that it would be weird to name this component FancyButton.js instead of FancyButton. What do you think? Moreover, in case component name was given I would remove duplication.

```jsx
// in FancyButton.js 
import styled from 'styled-components'
const FancyButton = styled.button` `
export default FancyButton
// will have displayName == 'FancyButton', not FancyButton.js__FancyButton  
```

`displayname` is not a word (filename - is), so I will rename filename to fileName for consistency.

mxstbr:
Yeah I guess you're right, this is fine.

Alright let's rename the option and then we can ship this 🔥 Good job!

PR #11, opened 2016-11-28
Import declaration detection enhancements

vdanchenkov:
Make importedVariableName value local to processed file.
Look for imports in separate traverse to catch hoisted imports.
Support `import {default as s} from ‘styled-components’`


mxstbr:
Huh, interesting, nice fix! LGTM

PR #13, opened 2016-12-04
Move code around, reduce output.

vdanchenkov:
@mxstbr take a note of changed output. It might require changes in styled-components code, but I doubt it.

As we are going to minify CSS, I'd start with this changes.


mxstbr:
Awesome, LGTM! No changes required 👍

PR #15, opened 2016-12-17
Minify css

vdanchenkov:


vdanchenkov:
I think, It's fine for now. More sophisticated optimizations could be added in the future.

src/index.js
```diff
@@ -18,23 +19,39 @@ export default function({ types: t }) {
...
+              path.get('specifiers').forEach((specifier) => {
+                let importedName
+                if (specifier.isImportDefaultSpecifier()) {
+                  importedName = 'default'
+                } else if (specifier.isImportSpecifier()) {
+                  importedName = specifier.node.imported.name
+                }
+                if (importedVariableNames.hasOwnProperty(importedName)) {
+                  importedVariableNames[importedName] = specifier.node.local.name
+                }
```
mxstbr: Can we extract this whole imports-stuff in a common module so we don't have the exact same code in the stylelint processor as in the babel plugin do you reckon? Or not worth the effort?

vdanchenkov: Nice idea. I think it worth the effort if it will be general enough to be used by 3-rd party plugins.

mxstbr: Alright, we should do that then!

PR #17, opened 2016-12-19
Switch to new API withConfig, fixes #16

### LINKED ISSUE #16: Update to new syntax
As of https://github.com/styled-components/styled-components/pull/227 the new syntax for the private API is `.withConfig`:

```JS
const Box = styled.div.withConfig({ displayName: 'Box', componentId: 'Button_asdf123' })`
  styles: here;
`
```

*This avoids issues on ReactNative where `target` would before need to be `View` (no string) and we'd need to import it, which would've made everything much more compilcated.*

I'm thinking about switching this to be `.__config` to make sure people don't accidentally use it, but that's a trivial change. We should get this changed asap!

vdanchenkov:


mxstbr:
`identifier` is now `componentId` (not sure why Glen renamed that...) 

Have you tried this on a project running on this branch https://github.com/styled-components/styled-components/pull/227?

vdanchenkov:
@mxstbr Fixed. And checked on the real project with styled-components/styled-components#227. Works fine.

PR #21, opened 2016-12-26
Transpile template literals

vdanchenkov:
Another feature for those who care about minification. @mxstbr Please put attention on docs and suggest/make changes if something there is difficult to comprehend.


mxstbr:
This looks great! Is there any reason to not enable this by default?

vdanchenkov:
@mxstbr This is something I was not sure about. The only reason to disable I can see is that transpilation will make bundle size slightly bigger given you don't care about transpilation at all. 

But you can always disable this feature in that rare case... Actually, I'm thinking about making it enabled by default as it's the case for most users.

mxstbr:
Yeah that sounds reasonable to me!

vdanchenkov:
Made it enabled by default and updated docs.

PR #22, opened 2017-01-11
Allow chains of member calls

vdanchenkov:
This change loosens checks to detect 
```
styled.div.attrs().anotherCall()`width: 100%;`
``` 
as styled component. I think it's a good change anyway, but besides that it allows syntax proposed in 
styled-components/styled-components#365.

### DIFF
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -14,24 +14,20 @@ const importLocalName = (name, state) => {
 }
 
 export const isStyled = (tag, state) => {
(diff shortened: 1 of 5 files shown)


mxstbr:
Let's merge this and `npm publish` a new version so people using the `v2` branch can start experimenting?

vdanchenkov:
@mxstbr Good idea. I want to merge another one tiny change. After that, could you take care of the release?

PR #25, opened 2017-01-26
[WIP] Preprocess CSS

philpl:
- [x] Minimal CSS parser that outputs an AST
- [x] Un-nesting transformer that flattens the AST
- [ ] Determine the format of the result we want to generate for `styled-components` to consume
- [ ] Add `preprocessTemplateLiterals` visitor (Very easy; can be seen in the `postcss-preprocessing` branch)
- [ ] Determine how preprocessed CSS in interpolations should work
- [ ] Write ludicrously high amount of unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "license": "MIT",
   "devDependencies": {
     "babel-cli": "^6.4.5",
+    "babel-jest": "^18.0.0",
     "babel-preset-env": "^1.1.8",
     "jest": "^17.0.3"
   },
@@ -31,5 +32,14 @@
     "server-side rendering",
     "ssr",
     "displayName"
-  ]
+  ],
+  "jest": {
+    "transform": {
+      ".js": "<rootDir>/node_modules/babel-jest"
+    },
+    "testEnvironment": "node",
+    "transformIgnorePatterns": [
+      "/node_modules/"
+    ]
+  }
 }
--- a/src/css/parser.js
+++ b/src/css/parser.js
@@ -1,0 +1,126 @@
+const parse = css => {
+  const findToken = (needle, start, end) => {
+    let index = start;
+
+    while (
+      css[index] !== needle &&
+      index <= end
+    ) {
+      index++;
+    }
+
+    // Throw if there's no matching token
+    if (index > end) {
+      throw new Error(`Unexpected token! Unable to find matching '${needle}'.`);
+    }
+
+    return index;
+  };
+
+  const findClosingParanthesis = (start, end) => {
+    let level = 1;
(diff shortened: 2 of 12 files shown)


philpl:
So for now I completed a minimal CSS parser and have a transformer that flattens it. To continue I'll need to figure out what shape the result should take on for styled-components to consume. Also I haven't quite figured out how the `keyframes` and `css` helpers should behave. Especially the `css` one might be tricky, since that will be another interpolation that might have to be combined with the higher level one.

It's still missing a lot of testing since this is pretty much just some experimental stuff right now.

mxstbr:
I reckon the shape needs to be some sort of object that can be parsed with very little effort. Possibly we could even put in just the AST? Would that be too big possibly?

philpl:
@mxstbr I do believe that is too big. I think we want to get as close to a normal string as possible, while being able to replace `&` in the selector values that contain it. Since we are already able to parse template strings, an object of the usual structure might be suitable?

Something like this:

```js
{
  '&': [[ 'color: ', ';' ], props => props.color ],
  '&:hover': [[ 'color: green;' ]]
}
```

philpl:
... Actually we can interleave this and diverge from what template strings need to do:

```js
{
  '&': [ 'color: ', props => props.color, ';' ],
  '&:hover': [ 'color: green;' ]
}
```

Since we don't need to differentiate between interpolations and CSS strings anymore, this should be fine, no?

mxstbr:
Yeah that sounds reasonable! @geelen what do you think?

philpl:
@mxstbr Just noticed a small problem with the selector un-nesting. 

```css
& {
  &:hover {
  }
}
```

The nested rule here is gonna be kept as `&:hover`.

```css
& {
  > div {
  }
}
```

While the nested rule here is going to be `& > div`... So if this selector is a function, like:

```js
`& {
  ${x => x ? '&:hover' : '> div'} {
  }
}`
```

Then we don't know whether to prepend the ancestor's selector or not :(

philpl:
It turns out that stylis.js is extremely forgiving with placeholders and really forgiving. So I might replace this with some stylis magic! ✨ We can still keep the AST parser around for when we decide to actually do static/dynamic css splitting, but for now it's totes not needed 😄

philpl:
Another thing 😭 inline placeholders that are directly before `{` are recognised as blocks of their own 👎 Maybe I was a bit fast to try to solve this with stylis

philpl:
Closing this PR in favour of #26. That one will probably be closer to how `v2` already processes its CSS, and it will add less surface area for bugs.

PR #26, opened 2017-01-28
Preprocess CSS using Stylis

philpl:
- [x] Stylis preprocessor with placeholders and a temporary classname
- [x] Visitor that passes the template string arguments into the preprocessor
- [x] Determine the format of the result we want to generate for `styled-components` to consume
- [x] Write ludicrously high amount of unit tests
- [x] ~Write preprocessor that works for `css` and `keyframes`~ (They should actually "just work™")
- [x] ~Disable minification when preprocessing is on~ (Separate PR)
- [x] Write simplified preprocessor for `keyframes` helper

This adds the transpilation step that preprocesses CSS, which is then accepted in styled-components under a new entrypoint. An open PR for accepting the preprocessed CSS is here: https://github.com/styled-components/styled-components/pull/585


mxstbr:
That looks amazing! 😍

One note: the `identifier` we pass through from the Babel plugin is the className that will later be assigned to the component, so we can even replace `&` with that!

mxstbr:
Is this done?!

philpl:
@mxstbr It's a bit rough around the edges :) The `keyframes` helper will need a simpler preprocessor, but that won't take long to write. Most of the development of this will need to be on `styled-components` itself now ;)

philpl:
@mxstbr The last commit adds the missing stuff for the keyframes, so this is complete, if we go by the original to-do list :tada:

I'm treading carefully here, because I spotted another limitation: You won't be able to add new rulesets inside your styled template strings, since it won't be flattened...

That was something we spotted earlier, but we need to resolve it now, I guess. To solve this, all interpolation-rulesets need to be wrapped in `css`. (Again sth we already wanted)

But, since the `css` helper is still wrapped inside a block, we will need to flatten it at runtime *urg*. For that we will need to gather the selector in which the `css` interpolation is in and append the new ruleset... somewhere. This is not as trivial as I thought, and I'm not sure whether we can make this problem simpler somehow?

If we want to do this at compile-time, we could recursively go through all interpolations and flatten them all into one `styled` tag, essentially removing all `css` helpers. But we'd run into the problem that a `css` interpolation can be passed in as a variable... And my babel-plugin-fu is not as good as I wish it were, to resolve that ^^

vdanchenkov:
Unfortunately, babel-plugin-fu have limits. It will not help to flatten css interpolation if it is imported from another file.  

```
import {verticalCenter} from 'helpers'

const Component = styled.div`
  ${verticalCenter}
`
```

Also, I want to ask if you considered conditionally applied rulesets? Something like.

```
import {verticalCenter, horizontalCenter} from 'helpers'

const Component = styled.div`
  ${props.direction == 'v' ? verticalCenter : horizontalCenter }
`
```

philpl:
@vdanchenkov Exactly these two things are the ones I'm concerned about.
I don't think we can do anything about them at compile-time.

That's where we gotta bring the magic to runtime ^^

philpl:
Since this is now behind a default-false flag and has an "experimental" note, I guess we can ship it :shipit:

wmertens:
@mxstbr I can't wait to test this out! npm publish?

mxstbr:
Test suite fails locally, so Ican't publish :confused:

philpl:
@mxstbr Oops! Seems like my stylis was out of date, when I updated the fixtures 😆  I updated it to 1.0.6 in the package.json

https://github.com/styled-components/babel-plugin-styled-components/commit/9fc0a6e

Issue #7, opened 2016-11-26
Add displayName to unnamed components exported as default, based on filename


vdanchenkov:
```jsx
// in Button.js
export default styled.div``
```
should have displayName == 'Button'

I'm going to implement it this weekend.

mxstbr:
Awesome, that's a great idea!

What about doing BEM style naming generally:

```JS
// shared.js

const Button = styled.button``

export { Button }
```

This `Button` could now have a `displayName` (and identifier, thus class) of `shared__Button-asdf123`. That way finding components becomes really really easy!

PR #27, opened 2017-01-30
[Proposal] Use shortids for componentId generation

philpl:
This uses the `displayName` and a `shortId` for generating a component id, to be as unique as possible, while still being human readable, if the user needs it.

It also removes the option to include filenames in the `displayName` (and thus the previous `componentId`), because we might not have it available, and it probably isn't much use to the user.


vdanchenkov:
filenames usage comes from #7 and #8. It's not for componentId, It's to add human-readable scope and make displayName BEM-like. Useful when you have a lot of components with the same variable name `const Container = styled.div` or do not have the name `export default styled.div`

PR #30, opened 2017-02-07
Workaround for line comments

### LINKED ISSUE #29: //-style comments are mangled by the plugin
When using line comments, they get their newline stripped by the minification and they become an erroneous selector.

<strike>I was also not able to turn off minify with the plugin options.</strike> (config error)

I propose that minification be disabled until #26 lands, that will be much more useful and correct.

vdanchenkov:
fixes #29

### DIFF
--- a/src/visitors/minify.js
+++ b/src/visitors/minify.js
@@ -2,12 +2,22 @@ import * as t from 'babel-types'
 import { useMinify } from '../utils/options'
 import { isStyled, isHelper } from '../utils/detectors'
 
+const minify = (linebreak) => {
+  const regex = new RegExp(linebreak + '\\s*', 'g')
+  return (code) => code.split(regex).filter(line => line.length > 0).map((line) => {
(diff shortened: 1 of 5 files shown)


kopax:
While waiting for this, is it normal that I still see CSS comments in my compressed project using this plugins ?

wmertens:
🎉 , but just curious why you don't just let stylis handle the minifying? Stylis supports line comments… (I'm probably missing something basic)

mxstbr:
> why you don't just let stylis handle the minifying?

Then you're shipping un-minified styles in your JS bundle which are massively huge, we have to do that at build time.

wmertens:
how about processing with stylis at build time as well? Replace the interpolations with known constants that you then split back on?

mxstbr:
That's what @philpl is working on in the open PR, but it might not be ready in time so this is a good intermediary solution.

PR #34, opened 2017-03-14
Disable SSR option by default

mxstbr:
Given our amazing ability to improve the debugging experience and to minify the styles, a lot of people who aren't server-side rendering will want to use this plugin, so I think it makes sense to disable SSR by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # `babel-plugin-styled-components`
 
-Babel plugin for `styled-components`. This is **only necessary if you're server-side rendering**, you can use `styled-components` perfectly fine without this Babel plugin. (it does give you a nicer debugging experience though)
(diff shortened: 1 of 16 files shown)


philpl:
Sounds reasonable! 👍 

But we'd want to remove this note then, I guess, and add a better description of all the things the plugin can do, right?

> ⚠️ This plugin is only necessary if you're server-side rendering, you can use `styled-components` perfectly fine without it! ⚠️

mxstbr:
That's the way it is though, it's only _necessary_ if you're doing SSR. I don't want people to think they can't use styled-components without this plugin!

philpl:
True, since the preprocessing is experimental that should suffice. I guess I was imagining it to do more than it actually does :laughing:

Let me fix the tests for you and merge this

mxstbr:
Thanks!

PR #37, opened 2017-04-03
Add 'module' option to use `no-parser.js` as an ES module

YasserKaddour:
Setting the `module` option to true will change `require('styled-components/no-parser')`
to `require('styled-components/dist/no-parser.es')`

Since bundlers can't detect the ES module, this should do the trick
What do you think?

Related to styled-components/styled-components#640

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,8 +85,7 @@ You can enable preprocessing with the `preprocess` option:
   ]
 }
 ```
-
-You can use the `no-parser.js` entry as an ES module by enabling the `module` option:
+If you use Rollup or webpack v2 you can take advantage of tree-shaking by setting the `module` option to true:
 
 ```JSON
 {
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -86,6 +86,19 @@ You can enable preprocessing with the `preprocess` option:
 }
 ```
 
+You can use the `no-parser.js` entry as an ES module by enabling the `module` option:
```
mxstbr: I can't say if this is a good solution, maybe @philpl has a better idea, but anyhow we should document this better since people won't have a clue what this means :laughing:

What about

> If you use Rollup or webpack v2 you can take advantage of tree-shaking by setting the `module` option to true.

Is that the case for the user? What other scenarios exist where somebody would want to use this?

YasserKaddour: Done, but for the other scenarios I have no clues

philpl:
@mxstbr I think it's a neat temporary solution, except if we can somehow find out whether babel is being run with the module transpilation enabled or disabled?

mxstbr:
Maybe @hzoo has an idea how we could do that?

PR #38, opened 2017-04-03
Fix test description

YasserKaddour:
I think the names of these two folder were swapped :wink:


philpl:
Oops! Yep, you're right

PR #40, opened 2017-04-10
Add static CSS extraction

philpl:
This is a proof of concept of parsing some static CSS rules and outputting them to a css file.
Currently this only works for styled components (`styled`) and none of the helpers.

To get this to work on the helpers (future PR?), we'll need some work on styled-components. Namely `css`, `keyframes`, and `injectGlobal` will need a method to inject an extra classname.


philpl:
I'm not really sure when to write all collected CSS to a file, and when the `Program` visitor is triggered. I think it's before and after a file, which is problematic, since we'll probably deal with multiple files all the time.

@thejameskyle to the rescue, plz? 😄

thejameskyle:
Build it as a webpack loader instead https://github.com/thejameskyle/dynamic-import-reporter-loader

Issue #39, opened 2017-04-04
Information about new warning in documentation


kopax:
> :warning: This plugin is only necessary if you're server-side rendering, you can use styled-components perfectly fine without it! :warning:

I was using plugin v1.0.0 to compress my distribution build. Is this included by default in styled-components@? without any plugin ?

mxstbr:
Nope, the warning says "is only _necessary_", not "should only be used when"! Any ideas how to improve that phrasing to make it clearer?

kopax:
Well, the warning is false in this case.

> :warning: This plugin is required if you're using server-side rendering and necessary if you require production distribution optimization, you can use styled-components perfectly fine without it! :warning:

mxstbr:
> it is also necessary for optimizing the production build

Not true, it's not _necessary_ in that case, that's exactly what I was trying to avoid saying! :wink: There's lots of people using `styled-components` in production without this plugin, and if we start saying that people will think they can't use `styled-components` with `create-react-app`.

I agree that we need to change the wording though, what about:

> ⚠️ This plugin is recommended to use with `styled-components` to get a smaller production bundle, but it's **only necessary if you're server-side rendering**. You can use styled-components perfectly fine without it! ⚠️

Issue #2, opened 2016-11-22
Disable displayName stuff in production


mxstbr:
We should check for `process.env.NODE_ENV !== 'development'` and disable `displayName` if that's the case.

vdanchenkov:
The issue was created when we used IIFE and had no configuration flags.

These days, I'd recommend using https://babeljs.io/docs/usage/babelrc/#env-option to configure it explicitly.

@mxstbr  WDYT?

mxstbr:
Sounds very reasonable!

breadadams:
In case it helps anyone else, the updated babel URL would be: https://babeljs.io/docs/en/config-files#apienv

Here's an example of disabling `displayName` for production:

```
// .babelrc.js

module.exports = function (api) {
  return {
    plugins: [
      ["styled-components", { ssr: true, displayName: !api.env("production") }]
    ]
  };
};
```

PR #46, opened 2017-04-26
Clarify message about plugin being needed for SSR

p-jackson:
Updated the docs after the discussion in #39.

Removed the overall bold styling so more specific bold styling could be done. I think the emoji's make it stand out enough any, unlikely it'll be missed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 This plugin adds support for server-side rendering, for minification of styles and gives you a nicer debugging experience when using `styled-components`.
 
-**⚠️ This plugin is only necessary if you're server-side rendering, you can use `styled-components` perfectly fine without it! ⚠️**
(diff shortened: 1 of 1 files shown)


evan-scott-zocdoc:
I have one additional question to ask in regards to the SSR feature that may be useful as a readme comment as well:

Does `ssr: true` need to be added to the client-side bundle config as well, or only the server-side bundle config? @philpl

So for instance, webpack bundle #1 has target: node for SSR and bundle #2 has target: web for browsers.

mxstbr:
It needs to be added to both!

mxstbr:
The documentation has been moved to the [website](https://github.com/styled-components/styled-components-website), mind submitting a PR over there?

p-jackson:
All good. I actually think the first paragraph of the new docs explains the plugin perfectly, so doesn't need this change.

Really like the new website by the way 👍

mxstbr:
Thank you for the kind words!

PR #54, opened 2017-06-03
[Component ID] Stabler ID by using mtime instead of code.

alloy:
Our specific situation is that the server-side source is being JIT compiled with `ts-babel-node`, whereas our client-side source is compiled and transformed through webpack. In the case of the client-side source, we add source map URLs to the output specifically for the browser to pickup.

This means that while the actual code is identical, the blob that this plugin receives is slightly different and thus the hash will be different, leading to the dreaded “React attempted to reuse markup in a container but the checksum was invalid.” issue.

I decided to replace the source code hashing by file mtime, which (when combined with filename) should be just as reliable, but allows for a bit more flexibility where code is semantically the same but the transformation results are slightly different.

### DIFF
--- a/src/utils/getMTime.js
+++ b/src/utils/getMTime.js
@@ -1,0 +1,3 @@
+import { statSync } from 'fs'
+
+export default filename => statSync(filename).mtime.getTime().toString()
--- a/README.md
+++ b/README.md
@@ -44,6 +44,34 @@ If you want server-side rendering support you can enable it with the `ssr` optio
 }
 ```
 
+By default this will use the `use-content` option, which will use the input code of the plugin to generate a hash from.
+
+If, for whatever reason, your pipelines differ between server and client side compilation there’s the possibility this
+leads to different hashes. In this case you can use the `use-mtime` option, which will use the file’s modification time
+to generate a hash from. It is not recommended to use this option in your test environment, as file modification times
+are not persisted across e.g. git clones. Luckily, Babel allows you to specify different options
+[based on the env](https://babeljs.io/docs/usage/babelrc/#env-option), which means you can set it to `use-content` in
+tests. The `.babelrc` configuration for would look like:
+
+```JSON
+{
+  "plugins": [
+    ["styled-components", {
+      "ssr": "use-mtime"
+    }]
+  ],
+  "env": {
+    "test": {
+      "plugins": [
+        ["styled-components", {
+          "ssr": "use-content"
+        }]
+      ]
+    }
+  }
+}
+```
+
 ### Better debugging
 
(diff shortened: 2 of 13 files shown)


src/utils/getMTime.js
```diff
@@ -0,0 +1,3 @@
+import { statSync } from 'fs'
+
+export default filename => statSync(filename).mtime.getTime().toString()
```
alloy: Decided to put this in its own module so that it would be easy to mock in this plugin‘s tests.

mxstbr:
Interesting interesting, I like it! Have you tested that this patch solves your problem?

/cc @vdanchenkov @philpl

alloy:
Yup, it does 👍

alloy:
Interesting, for some reason this is failing on our CI https://github.com/artsy/reaction-force/pull/168. However it does work in practice and tests locally are green, nonetheless I’ll take a look into what the issue is later today.

philpl:
@mxstbr @alloy Very clever idea! I'll look into it more and think about it a little, but there shouldn't be any edge cases, I believe.

alloy:
Hmm, one possible issue I just thought of is that iirc git doesn't restore mtime, which leads to different results on different machines. This is not an issue in general, except during test runs where the hash may have been recorded (e.g. jest snapshots).

Sigh. Maybe the plugin should just read the _original_ file and hash its contents instead (which would be the TypeScript file in our case)?

mxstbr:
> Maybe the plugin should just read the original file and hash its contents instead (which would be the TypeScript file in our case)?

That could be an idea, but would make the plugin very slow, right?

alloy:
Yeah indeed. I just pushed another commit that allows you to choose what type of seed to use (defaults to the previous content based version) which makes it possible to use the content based approach in tests.

I’m not a big fan of more options, but seeing as this is something you configure once rather than o each CI invocation, I think it’s not that bad.

alloy:
I should amend the README to say that this isn't recommended to use in the test environment _if you're recording the hash, e.g. in jest snapshots_.

philpl:
@alloy I don't think this is a good option tbh :/ There's a lot to consider here and I'd rather come up with something that also works well, once we integrate this stuff into webpack. It's problematic to use the original file contents for hashing purposes obviously, and multiple options could end up confusing people.

I don't think we'd like to confuse people on a plugin that will end up being "less optional" in the future ;)

alloy:
Yeah that's fair. I should add that my specific case is solved by just removing the source map URL. Would you prefer to start by normalizing by getting rid of that from the input?

alloy:
Ok I ended up fixing our specific issue elsewhere, because it appears that some compilation phase was accidentally leaving in the source map URL. I think there’s still room for thinking about how this could be made more fool proof, but for now I’ll close this specific PR and leave it up to you to decide what/how to go forward.

PR #55, opened 2017-06-11
Strip comments properly from styled template strings

philpl:
Fixes: https://github.com/styled-components/styled-components/issues/888

The new logic works like this:

- Joins quasis with placeholders
- Removes multi-line comments
- Splits by newlines
- Removes line-comments
- Filters lines with only line-comments
- Joins

This thus removes all comments and gives us greater
granularity for minification, since it's only a single
string we now operate on. This should prevent mangling
in the future.

### DIFF
--- a/src/visitors/minify.js
+++ b/src/visitors/minify.js
@@ -4,9 +4,7 @@ import { useMinify, useCSSPreprocessor } from '../utils/options'
 import { isStyled, isHelper } from '../utils/detectors'
 import { makePlaceholder, splitByPlaceholders } from '../css/placeholderUtils'
 
-const newline = newline => new RegExp(newline + '\\s*', 'g')
-const multilineComment = newline => new RegExp('\\/\\*(.|' + newline + ')*?\\*\\/', 'g')
-
+const makeMultilineCommentRegex = newlinePattern => new RegExp('\\/\\*(.|' + newlinePattern + ')*?\\*\\/', 'g')
 const lineCommentStart = /\/\//g
 
 // Counts occurences of substr inside str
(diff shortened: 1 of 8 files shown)


philpl:
@kopax Since we had some reports over at the SC repo of the minify option not working correctly, we had to actually implement this now. We should've done this earlier, as you requested, I guess :wink:

thysultan:
@philpl If you are running this through stylis, would it make sense to delegate the removal of comments to stylis since it handles removing both line and block comments?

thysultan:
^^ and use the `stylis.set({compress: true})` option to minify further.

philpl:
@thysultan We're actually not running this through stylis. The `minify` option is disabled when `preprocess` is turned on. And we're using stylis for `preprocess` only.

Flydiverny:
@philpl Re-enabling minify with this PR linked in seems to fix the issues I had in styled-components/styled-components#888 👍

src/visitors/minify.js
```diff
@@ -1,19 +1,77 @@
...
-  useMinify,
-  useCSSPreprocessor
-} from '../utils/options'
+
+import { useMinify, useCSSPreprocessor } from '../utils/options'
 import { isStyled, isHelper } from '../utils/detectors'
+import { makePlaceholder, splitByPlaceholders } from '../css/placeholderUtils'
+
+const newline = newline => new RegExp(newline + '\\s*', 'g')
+const multilineComment = newline => new RegExp('\\/\\*(.|' + newline + ')*?\\*\\/', 'g')
```
mxstbr: Nit: fix naming `newline` x4?

philpl: I'll replace it with something more readable :)

philpl:
@mxstbr sure! Do you think we should test the helper functions atomically? Also, should we add more fixtures? I couldn't come up with more on the spot

philpl:
@mxstbr should be fine now 😄 wanna take another look?

philpl:
@mxstbr Cheers! I'll test it some more and release the patched version tonight :smile:

evan-scott-zocdoc:
Just ran into this today - glad there's a fix!

kopax:
May we have a release including this ?

PR #58, opened 2017-06-14
Upgrading stylis to match version used by styled-components

silverlight513:
- Upgraded stylis version in package.json
 - Upgraded yarn.lock to match package.json
 - Upgraded whitespace in test results to match updated stylis output

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,6 @@
     "displayName"
   ],
   "dependencies": {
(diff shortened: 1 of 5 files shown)


philpl:
hm, I thought this would mess with the preprocessing, but seeing that it doesn't I'll see tomorrow whether everything is in order. Cheers! :)

PR #70, opened 2017-06-27
Fix newlines being mangled inside values

philpl:


chrisvoll:
@philpl I just tested this with our code and can confirm that it fixed our issue. Thanks so much for the quick turnaround!

mxstbr:
Want to publish a patch v?

philpl:
@mxstbr will do :)

PR #72, opened 2017-07-01
Upgrade to stylis 3.2 and use constructor

philpl:
This prevents bugs with shared options and middleware when different
babel plugins use stylis.

cc @thysultan


thysultan:
LGTM 👍

Issue #73, opened 2017-07-10
Component display name appears twice in single CSS class


andrewtpoe:
When using `babel-plugin-styled-components`, the display names in the identifying CSS classes are being doubled up. It appears this problem may actually go deeper and be caused by the display names themselves being doubled.

This screenshot is from the React dev tools of [a very simple app](https://andrewtpoe.github.io/sc-babel-bug/) I put together to demonstrate this problem. You can find the code [here](https://github.com/andrewtpoe/sc-babel-bug).

![screen shot 2017-07-10 at 5 54 48 pm](https://user-images.githubusercontent.com/11183459/28041857-eae8dd44-6598-11e7-9dba-17a90b1a1b13.png)

I now have 3 sites where I am attempting to use this plugin and all behave similarly. The primary site uses a custom webpack config, the others use create-react-app.

Thanks for having a look at this. We are very excited to have the display names in the CSS (once...).

mxstbr:
@philpl any ideas why this is happening? I'm assuming it's not intended.

andrewtpoe:
Definitely not intended. I don't have any ideas concerning it yet. It appears to happen even with the most basic config.

philpl:
It is actually intentional. The `displayName` includes both the guessed component name and the filename: https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/visitors/displayNameAndId.js#L39

andrewtpoe:
Is there a way to bypass that behavior? It is very unexpected and undesired.

philpl:
@mxstbr that sounds like a good solution here. Shouldn't be hard to add :) Any takers? Otherwise I'll do it when I have time

probablyup:
@philpl is this fixed? https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/visitors/displayNameAndId.js#L40

mxstbr:
Yes, thank you!

PR #74, opened 2017-07-11
Resolves double component name in display name

andrewtpoe:
This PR resolves the [bug](https://github.com/styled-components/babel-plugin-styled-components/issues/73) where the `displayName` and CSS class name include duplicates of the same name when the `componentName` is identical to the `blockName`.


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.7",
```
philpl: Can you remove the version bump please? We'll take care of it when publishing ;)

andrewtpoe:
I've removed the version bump. If it's all the same, I'll leave the squash to you when merging. 😄

test/fixtures/05-use-file-name/before.js
```diff
@@ -1,5 +1,6 @@
 import styled from "styled-components";
 
 const Test = styled.div`color: red;`;
+const before = styled.div`color: blue;`;
```
philpl: Hm, sorry, one more thing ^^"

Can you add an "inverted" test here? A component that doesn't have the name `before`? This way we'll have both cases covered in this file

andrewtpoe: I'm not sure I follow exactly. Isn't that what `Test` is directly above this component?

philpl: Oh true! Sorry, I must've overlooked it

PR #83, opened 2017-08-20
Sanitize display name to prevent invalid class names

### LINKED ISSUE #80: displayName and filenames with dots
If the filename for your styled-component is e.g. countdown.styled.js you get a classes like "countdown.styled__Item-xwcjw2-1 ellRMD". Since this contains a dot it breaks references to that component.

Example:
```
export const Item = styled.div`
  display: inline-block;
`

export const Num = styled.div`
  border-right: 1px solid red;
  
  ${Item}:last-child & {
    border-right: 0;
  }
`

const Countdown = () => <div>
  <Item><Num>1</Num> day</Item>
  <Item><Num>2</Num> hours</Item>
  <Item><Num>3</Num> minutes</Item>
</div>
```

The example works with displayName: false but not with displayName: true.
Removing the dot from the filename also fixes the problem.

I suggest stripping dots and other reserved characters from the classnames.

niklasgrahl:
Fixes #80

It removes invalid characters from the display name so it can be used as a valid classname.

### DIFF
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -94,9 +94,11 @@ const getComponentId = (state) => {
 
 export default (path, state) => {
   if (isStyled(path.node.tag, state)) {
+    const displayName = useDisplayName(state) && getDisplayName(path, useFileName(state) && state);
+
     addConfig(
       path,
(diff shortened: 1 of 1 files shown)


niklasgrahl:
Sure, changed it!

PR #84, opened 2017-08-25
Use file.opts.filename instead of file.opts.basename for babel 7 compatibility

lixiaoyan:
https://babeljs.io/blog/2017/03/01/upgrade-to-babel-7-for-tool-authors

### DIFF
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -27,9 +27,8 @@ const addConfig = (path, displayName, componentId) => {
 }
 
 const getBlockName = (file) => {
-  return file.opts.basename !== 'index' ?
-    file.opts.basename :
-    path.basename(path.dirname(file.opts.filename))
+  const name = path.basename(file.opts.filename, path.extname(file.opts.filename))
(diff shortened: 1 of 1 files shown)


lixiaoyan:
@JamieDixon Seems both filename and basename are available in Babel 6, but in Babel 7 only filename.

mxstbr:
It seems they are different things though? The changelog says 

> The opts.basename option exposed on state.file.opts has been removed. **If you need it, best to build it from opts.filename yourself**

But it doesn't say how to do that 🙃 /cc @loganfsmyth maybe has some wisdom to drop here?

lixiaoyan:
https://github.com/babel/babel/pull/5467/files#diff-d6903991e1c21aaacc4d615157f587e5L144
Seems babel use a different way to get basename but I'm sure they are same.
Anyway, I'll update my code to follow the way babel used.

loganfsmyth:
Good call, thanks for cleaning this up. This option was never documented anywhere and it just didn't seem like something worth having since you can do this to get it from the filename.

lixiaoyan:
Could you make a release for this, thanks.

mxstbr:
Of course, my bad!

PR #86, opened 2017-09-10
Failing test with react hot loader

piteer1:


test/fixtures/19-display-correct-name-with-default-export-and-hot-loader/before.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import styled from 'styled-components';
+
+class Component extends React.Component {}
+
+export default styled(Component)`
```
philpl: This is not a recommended pattern anyway, so there's not much we can do. This seems to be caused by the order in which the plugins are run. The other way around it might work better.

For a default component you still should declare a variable name. The same goes for functional components in React. Otherwise the function, and in the plugin's case the `StyledComponent`, won't have a correct name.

This only works in this case because we attach the file name to it.

So overall, I'm pretty sure this is expected behaviour we can't fix, and it'll take modifications to your code. Have you tried what happens when you reverse the order of the plugins?

piteer1: @philpl yes I've tried to reverse order of plugins also and it finishes with the same result. 

As you said, I'm not sure if it should be "fixed" as this plugin works ok and you never know if there won't be any changes in either hot loader or some other plugin which changes the code. Additionally as you wrote, it works by simply doing:

```js
const StyledComponent = styled(Component)``;
export default StyledComponent;
```

and I've switched to this approach. Maybe the best fix is just putting it in the docs FAQ section (or something like similar) info about this problem when using `ssr`?

piteer1: @philpl should I close this PR?

PR #91, opened 2017-10-06
Remove stray .orig files

fatfisz:
Another small thing: I found some .orig files originating from this commit: 56db882059468c9af278f6e01a8ef81eff331d5e, probably a result of conflicts. This PR removes them all :)


philpl:
Shall we add those to the local gitignore? I've got them in my global one, but things tend to go wrong quite frequently when dealing with manual merges 😆

Issue #79, opened 2017-08-18
SSR with styled.extend throws error


srghma:
[minimal example (its with-styled-component next example)](https://mega.nz/#!o8cESJaC!hy319-Vn5VIfpHeJSteHlR9n0kWFjRMDfevEVAdXbm4)

What I do
1. yarn run dev
2. load page
3. reload page

Get error

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) ="1"><div class="sc-bdVaJa iNGvMq" data-
 (server) ="1"><div class="sc-jzJRlG fzsTbP" data-

```

mxstbr:
Thanks for the reproduction, appreciated!

BjornMelgaard:
@mxstbr will do whatever to get this fixed ;)

hmontes:
Same error using this tutorial:

https://medium.com/styled-components/building-a-blog-with-next-js-359cf1236574

nialloc9:
Same issue here.

fatfisz:
I'd like to tackle this, but first I have to "fix" all the other tests - I'm working on Windows and maybe that somehow contributes to the fact that the ids are different in every case.

<details>

    @@ -1,26 +1,26 @@
     import s from "styled-components";

     const Test = s.div.withConfig({
       displayName: "Test",
    -  componentId: "syqrfc-0"
    +  componentId: "pe3280-0"
     })`width:100%;`;
     const Test2 = true ? s.div.withConfig({
       displayName: "Test2",
    -  componentId: "syqrfc-1"
    +  componentId: "pe3280-1"
     })`` : s.div.withConfig({
       displayName: "Test2",
    -  componentId: "syqrfc-2"
    +  componentId: "pe3280-2"
     })``;
     const styles = { One: s.div.withConfig({
         displayName: "One",
    -    componentId: "syqrfc-3"
    +    componentId: "pe3280-3"
       })`` };
     let Component;
     Component = s.div.withConfig({
       displayName: "Component",
    -  componentId: "syqrfc-4"
    +  componentId: "pe3280-4"
     })``;
     const WrappedComponent = s(Inner).withConfig({
       displayName: "WrappedComponent",
    -  componentId: "syqrfc-5"
[...]

PR #92, opened 2017-10-06
Use RegExp instead of a string in String.prototype.replace

fatfisz:
As described in #79, `String.prototype.replace` needs at least a `RegExp` with the `global` flag to replace all occurrences of a text in the source text.

This concrete example was causing the tests to fail on Windows (only the first backslash was replaced with a slash).


philpl:
:heart:

PR #94, opened 2017-10-12
Fix #93 support require(), detect default call expression 

### LINKED ISSUE #93: Support require() calls
## Plugin will not work properly with Typescript

from https://github.com/styled-components/styled-components/issues/1168#issuecomment-335465598 (Include workaround)

`import` default in Babel and Typescript have a little different. The result of Typescript come with `.default` 

## Reproduce
1. Clone this repo https://github.com/zapkub/nextjs-typescript-styled-component
2. Modify `index.tsx` line 2 from `StyledComponents` to `styled` and remove line no. 3
3. Run project
4. Plugin will be broken

## Detail
babel-plugin-styled-components@1.2.1
styled-component@2.2.1
next@4.0.1

zapkub:
from #93 
# Support require()
Add detect styled variable on typescript `require` by finding styled `require` variable

ref
https://github.com/styled-components/styled-components/issues/1168#issuecomment-335465598


philpl:
cc @mxstbr wanna have a quick look?

src/utils/detectors.js
```diff
@@ -14,13 +14,29 @@ const importLocalName = (name, state) => {
...
     // styled.something()
     return isStyled(tag.callee.object, state)
   } else {
     return (
       (t.isMemberExpression(tag) && tag.object.name === importLocalName('default', state)) ||
-      (t.isCallExpression(tag) && tag.callee.name === importLocalName('default', state))
+      (t.isCallExpression(tag) && tag.callee.name === importLocalName('default', state)) ||
+
+      /**
+       * Support Typescript import #93
```
philpl: Small nit: I don't think we have to refer to the specific issue here. Also we do not need to refer to TypeScript at all. Multiple transpilers and other tools might change the import name. Even the user might change it on purpose and use `require()`. So I think we can keep the comment generic and say:

> styled-components might be imported using a `require()` call and assigned to a variable of any name.

The same we could add to `importLocalName` just using the default name

zapkub: I see the point, this issue cause from js require() itself not Typescript

zapkub:
any update for this pull request ?

PR #95, opened 2017-10-16
Update stylis dependency to 3.x

theshortcut:
Updates to stylis 3.3 and should help address issues like https://github.com/styled-components/styled-components/issues/1218 in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,6 @@
   ],
   "dependencies": {
     "babel-types": "^6.26.0",
-    "stylis": "^3.2.1"
+    "stylis": "3.x"
   }
 }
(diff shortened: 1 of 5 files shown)


jaydenseric:
[Deja vu](https://github.com/jaydenseric/apollo-upload-client/pull/12#issuecomment-305129994) 🙂

It might be a good idea to remove the Yarn lockfile (no package-lock.json?) if you explicitly don't want to pin a dependency. It is [best practice](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514) not to use lockfiles for packages anyway.

mxstbr:
Oh crap :laughing:

PR #104, opened 2017-11-06
deps: upgrade stylis, fix semver range notation

evan-scott-zocdoc:
I think there was something about the "3.x" syntax that was causing
this subdependency from being  automatically updated externally.

In my own projects, when running npm update I have not been getting
automatic stylis updates as they are released.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,6 @@
   ],
   "dependencies": {
     "babel-types": "^6.26.0",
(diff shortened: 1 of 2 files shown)


evan-scott-zocdoc:
Same change as https://github.com/styled-components/styled-components/pull/1296

probablyup:
Going to redo this under my personal account

PR #111, opened 2017-12-12
feat: add url prefix option

timkendall:
## What

This adds support for a `urlPrefix` option. When supplied, the url prefix will be added to all paths inside of `url(...)` declarations defined by styled-components. It's like Webpacks `baseUrl` for your styled-components!

For example, `url('/images/foo.png')` would become `url('https://bar.com/images/foo.png')`.

## Why

To support the serving of assets from a different domain than that serving the app.

## Other

Would be great if you guys want to support this. Our team has found it to be useful with our current deployment architecture. Also, this is my first time working with Babel plugins so I wouldn't be surprised if I missed something.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 *.log
 lib
+.DS_Store
--- a/src/utils/options.js
+++ b/src/utils/options.js
@@ -8,3 +8,4 @@ export const useFileName = (state) =>getOption(state, 'fileName')
 export const useMinify = (state) => getOption(state, 'minify')
 export const useCSSPreprocessor = (state) => getOption(state, 'preprocess', false) // EXPERIMENTAL
 export const useTranspileTemplateLiterals = (state) => getOption(state, 'transpileTemplateLiterals')
+export const useURLPrefix = (state) => getOption(state, 'urlPrefix', false)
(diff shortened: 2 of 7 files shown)


philpl:
This is awesome work! I do see great use for it, but I’m unsure whether it’s a concern that shall become part of this plugin. Since this plugin is mainly a part of styled-components and it’s functionality there are two questions that arise:

- wouldn’t this work and shouldn’t it be agnostic to any css-in-is library that uses tagged template literals?
- should it be part of a plugin that is concerned about some core functionalities of styled-components, e.g. debugging, SSR, CSS minification...

My opinion is that this would make an amazing plugin, but maybe it should be separated? I’m happy to answer any questions that you have to help you branch this out, if that’s the path we’ll be going down.

Let’s see what @mxstbr thinks :) thanks again for the contribution! I want to make sure that it ends up being usable, but let’s see if this is truly the right place for it. Hope you understand :)

timkendall:
Thanks! I agree and these considerations went through my mind as well. Turns out that our codebase does have a fair number of relative assets paths outside of styled-components as well. A separate Babel plugin would indeed be very useful.

That being said, I think including this functionality here may still be useful for users that only need to handle asset URLs declared in styled-components.

In any case, up to you guys ;)

mxstbr:
This definitely sounds awesome as a plugin, but I don't think it should live in the styled-components one—maybe make it its own package?

timkendall:
Works for me 👍 https://github.com/timkendall/babel-plugin-base-path is where I'll be implementing it. Thanks for all of your guys work on styled-components!

PR #112, opened 2017-12-20
use a simpler hashing algorithm when a file system is available 

### LINKED ISSUE #105: File hash in displayName is too fragile
Using a hash of the file contents when generating the displayName causes snapshot tests to fail frequently and unnecessarily. Changes to my files that don't effect the component require me to update my snapshots - the false negatives kind of make them redundant.

I assume the reason for the file hash is to make the class name unique, as 2 components could possibly have the same name.

Would it not make more sense, and be less fragile, to use a hash of the file path?

probablyup:
fixes #105

By not including the file contents in the hash, the generated IDs
should be more durable and less susceptible to minor differences
in file content due to a babel plugin being applied to ssr vs
client config or something like that


probablyup:
Going to add a snapshot serializer to do the whitespace removal stuff

probablyup:
Hmm. Any idea why the inner whitespace in the outputted CSS rules would be different @mxstbr?

probablyup:
@mxstbr even with the separate commits? I'll merge the whitespace serializer back into the first two commits to make it as clear as possible

probablyup:
There we go.

mxstbr:
Shipping this and the ref styled-components PR! https://github.com/styled-components/styled-components/pull/1381

PR #117, opened 2018-01-11
Add Uglify Pure Comments

osamajandali:
- Install babel-helper-annotate-as-pure that helped to add leading pure
comments
- Add an option called useUglifyPure which give the user the option to
disable it, and it is set true by default.
- By passing the path it automatically adds pure comments
- Snapshots is not updated to help out by checking the difference while
testing


package.json
```diff
@@ -34,6 +34,7 @@
     "displayName"
   ],
   "dependencies": {
+    "babel-helper-annotate-as-pure": "^7.0.0-beta.2",
```
existentialism: You'll want to use `@babel/helper-annotate-as-pure` (we switched to scoped packages)

osamajandali: should I update npm package for bable/types also, as now it is not using scoped packages yet

existentialism: To @Andarist's point, we're sneaking in a babel 7 package here and leveraging the fact that it's compatible with Babel 6. I don't think we want to switch to `@babel-types` here yet.

Sidenote, we can probably remove a dep on `babel-types` completely by passing [types down from here](https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/index.js#L7) also.

src/visitors/templateLiterals/index.js
```diff
@@ -14,4 +16,9 @@ export default (path, state) => {
   } else if (useTranspileTemplateLiterals(state)) {
     transpile(path, state)
   }
+
+  // add uglify pure (/*#__PURE__*/) comment before function calls
+  if(useUglifyPure(state)){
```
Andarist: im not sure when but sometimes `path.node.leadingComments` might be null, it would probably be good to ensure the node has this as array before calling `annotateAsPure`, simple:
```js
path.node.leadingComments = path.node.leadingComments || []
annotateAsPure(path)

```
should do the trick

osamajandali:
@Andarist Thank you very much man, I really appreciate your help.
I didn't get this 

> I also think that this plugin could add this comment for a lot more calls, i.e. existing styled

I see that it is added to styled calls, where is the point exactly.

Andarist:
Comments are tricky for get right in AST manipulation, it seems (but im not sure) that because you annotate `TemplateLiteral` (according to the file path - `src/visitors/templateLiterals/index.js`) comments are placed later in the output at "correct" positions, but it might be kinda an accident.

Also actually including updated fixtures would make IMHO reviewing this easier.

osamajandali:
Are you sure that this 
```javascript
styled.div('')
```
should have pure function. It is not being transpiled by styled component 
the input and the output of the about line on code is the same
```javascript
const Test4 = styled.div('');
```
Instead if used `styled.div``;` the output will be
```javascript
const Test4 = /*#__PURE__*/styled.div.withConfig({
    +  displayName: 'Test4'
    +})``;
```

probablyup:
Yeah it should be

Andarist:
@osamajandali yeah, that's the point - this plugin targets known usage of `styled` and it knows that it's safe to annotate those calls, so it should annotate them, even if this code is not transpiled at the moment in any other way

osamajandali:
I get the point and I manage to fix this issue.
But by fixing it, there is another thing now. When the function is called more than one time there will be  more than one comment.
For example:
```javascript
const Test2 = styled('div')``;
```
Will be
```javascript
+const Test2 = /*#__PURE__*/ /*#__PURE__*/styled('div').withConfig({
       displayName: 'Test2'
     })``;
```
I think this make sense but do you guys think that is not redundant.

Another thing do we have to add pure comments for requires?
For example:
```javascript
var _styledComponents = require('styled-components');
```
Will be
```javascript
var _styledComponents = /*#__PURE__*/require('styled-components');
```

Andarist:
@osamajandali I think that this is because you annotate CallExpression and TemplateLiteral, when the output code is later generated the leadingComment on TemplateLiteral gets moved to before this CallExpression - as mentioned comments are tricky for ASTs and some cases might not work "properly". Imagine this:
```js
styled/*foo*/`color: #fff;`
```
is this comment a leadingComment for TemplateLiteral or is it a trailingComment for CallExpression?

Because of that generated comments might get misplaced etc.

Anyway - you should focus on annotating CallExpressions only, TemplateLiterals do not need them. So unfortunately you'd actually have to refactor your PR.

osamajandali:
I annotate TemplateLiterals in the first place because it doesn't duplicate the comment. While annotating CallExpressions made the issue I was talking about.

osamajandali:
@mxstbr I will do this and I will made this option false by default for now.

osamajandali:
For more elaboration on multiple comment generated when annotating **CallExpression** 
Imagine a line of code from fixtures:
```javascript
const WrappedComponent = styled(Inner)``;
```
it will be annotated as follows when annotating inside CallExpression
```javascript
const WrappedComponent = /*#__PURE__*/ /*#__PURE__*/styled(Inner).withConfig({
      displayName: "WrappedComponent"
})``;
```
as you can see there are two pure comments.

This happen because we call **CallExpression** two times with different node which lead to add **leadingComments** in two different nodes.

The first node is:
 ```
{ type: 'CallExpression',
        callee:
         { type: 'MemberExpression',
           object:
            Node {
              type: 'CallExpression',
              start: 274,
              end: 287,
              loc: [Object],
              callee: [Object],
              arguments: [Array] },
           property: { type: 'Identifier', name: 'withConfig' },
           computed: false },
        arguments: [ { type: 'ObjectExpression', properties: [Array] } ] }
```
and the second one is:
```
Node {
        type: 'CallExpression',
        start: 274,
        end: 287,
        loc:
         SourceLocation {
           start: Position { line: 8, column: 25 },
           end: Position { line: 8, column: 38 } },
        callee:
         Node {
           type: 'Identifier',
           start: 274,
           end: 280,
           loc: SourceLocation { start: [Object], end: [Object], identifierName: 'styled' },
           name: 'styled' },
        arguments:
         [ Node {
             type: 'Identifier',
             start: 281,
             end: 286,
             loc: [Object],
             name: 'Inner' } ] }
```

While annotating in **TaggedTemplateExpression** doesn't do this duplication but will rise again the issue that this
```javascript
styled.div('')
```
will not have pure comments

Andarist:
You just have to bail out on nested CallExpressions, let's look at this:
```js
styled(Inner).withConfig({ displayName: "WrappedComponent" })
```

We have 2 CallExpressions here:
- first (outermost) which has a `callee` of `styled(Inner).withConfig`
- second (inner) which has a `callee` of `styled`

When annotating the inner `callee` you are supposed to get its outermost CallExpression and try to annotate it instead of annotating the inner one (or just skip annotating inner altogether if it leads to the same results regarding dropping this statement).

mxstbr:
Just needs some tests and then we should be good to go!

osamajandali:
@mxstbr just some question and we will be ready to go
- Should I make new fixture for this ?
- Should we make it active by default or leave it's activation to the users?
- Will we add this to documentation ?

mxstbr:
1. Yes, please make new fixtures for this feature specifically
2. It should be turned on by default, but please disable it manually for all existing test suites in their `.babelrc`s
3. Yes, an addition to the documentation would be awesome but I'm happy to do that once all the code is done :blush:

src/index.js
```diff
@@ -11,6 +14,11 @@ export default function({ types: t }) {
         noParserImportDeclaration(path, state)
       },
       CallExpression(path, state) {
+        if(useUglifyPure(state)){
+          if(path.parent.type == 'VariableDeclarator' || path.parent.type == 'TaggedTemplateExpression'){
```
Andarist: this one seems to annotate **every** `CallExpression`, or Im not aware how this plugin gets activated - I havent studied the whole source, but from this piece alone it seems that it will annotate everything, ie. `myImpureGlobal()`

osamajandali: Indeed yeah you will annotate everything but **once**.
As we were discussing before the issue of multiple pure comments were added, I fixed it by making sure that it is the outermost CallExpression only not the inner ones.

osamajandali: If you are talking about that it would annotate undesired annotation so I haven't found this problem while testing, but if you have a specific test that I could test against to double check please give it to me.

Andarist: the point is that this plugin should only "touch" `styled-components` related code, because it is the only code about which we can make any kind of assumptions

osamajandali: I get it now I will work on that.

src/visitors/uglifyPure.js
```diff
@@ -0,0 +1,19 @@
+import annotateAsPure from "@babel/helper-annotate-as-pure"
+
+import { useUglifyPure } from '../utils/options'
+import { isStyled, isHelper } from '../utils/detectors'
+
+export default (path, state) => {
```
Andarist: much better 💪

osamajandali: You helped me a lot big thanks to you 😁

osamajandali:
@mxstbr thanks for giving me this opportunity, I am very happy that my first PR is something related to styled component 😍

osamajandali:
Yeah sure, I will do it tonight.

src/visitors/uglifyPure.js
```diff
@@ -1,16 +1,18 @@
...
-import { isStyled, isHelper } from '../utils/detectors'
+import { isStyled, isHelper, isInjectGlobalHelper } from '../utils/detectors'
 
 export default (path, state) => {
     if(useUglifyPure(state)){
         if(isStyled(path.node,state) || 
           isStyled(path.node.callee,state) || 
-          isHelper(path.node.callee,state) 
+          isHelper(path.node.callee,state) ||
+          isInjectGlobalHelper(path.node.callee,state)
```
Andarist: i dont think we should annotate `injectGlobal`, its clearly effectful and removing calls to it will break most apps

osamajandali: I was thinking the same. @philpl why was you thinking we should annotate injectGlobal?

mxstbr: Yeah let's not annotate `injectGlobal` it is not pure 😅

philpl: True! My mistake :) let’s go ahead with this as it supports every other helper now

mxstbr:
Thank you so much for helping us improve styled-components! Based on our [Community Guidelines](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md) every person that has a PR of any kind merged is offered **an invitation to the styled-components organization**—that is you right now!

Should you accept, you'll get write access to the main repository. (and a fancy `styled-components` logo on your GitHub profile!) You'll be able to label and close issues, merge pull requests, create new branches, etc. We want you to help us out with those things, so don't be scared to do them! Make sure to read our [contribution and community guidelines](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md), which explains all of this in more detail. Of course, if you have any questions just let me know!

osamajandali:
@mxstbr Thank you very much, this totally made my day 😁 and I will be very active the next period..prepare for it 😄

PR #122, opened 2018-01-22
Module names

gabiseabra:
Adds "moduleName" configuration option.

### DIFF
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -1,14 +1,17 @@
 import * as t from 'babel-types'
+import { getModuleName } from "./options"
 
 const importLocalName = (name, state) => {
   let localName = name === 'default' ? 'styled' : name
+  let moduleNames = getModuleName(state)
+  if(typeof moduleNames === "string") moduleNames  = [ moduleNames ]
 
   state.file.path.traverse({
     ImportDeclaration: {
       exit(path) {
         const { node } = path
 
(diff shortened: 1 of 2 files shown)


bartels:
A benefit of making this configurable would be for users of other libraries that wrap the styled-components api. For instance, the xstyled library exports a customized version of `styled` that wraps the styled-components one. 

The babel plugin is not currently working for users of xstyled. I think It should work fine, except that looking through the code, I see  `VALID_TOP_LEVEL_IMPORT_PATHS`  hard-codes the valid module paths. So the plugin skips over ones like `import styled from '@xstyled/styled-components'` since it isn't in the styled-component module.

If this were user configurable in .babelrc, then xstyled users could get it working. 

The xstyled bug report here:  https://github.com/smooth-code/xstyled/issues/94

gregberge:
As the author of [xstyled](https://xstyled.dev/), I confirm that this would be awesome to be able to use the styled-components babel plugin without having to duplicate it.

agriffis:
I see now that this has conflicts, but does anybody have insight into what prevented this from being merged originally? It would be a huge help to users of xstyled

SpicyPete:
I would also like to see this PR make it through.  
Using Xstyled, and having legible class names would be great.

gregberge:
Already possible https://xstyled.dev/docs/migrate-from-styled-components/#babel-plugin

PR #128, opened 2018-02-22
Add Babel 7

damassi:
Swaps out Babel 6 with Babel 7, and updates Jest to latest.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    ["env", {
+    ["@babel/env", {
       "targets": {
         "node": 4
       }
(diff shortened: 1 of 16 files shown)


probablyup:
I'd like to wait on this until Babel 7 final is out

existentialism:
Beyond changing the build deps, one other option is to _remove_ the hard dependency on `babel-types`/`@babel-types`, and instead pass the types that are passed directly to the plugin to the util functions, as I mentioned back when we [added pure comments](https://github.com/styled-components/babel-plugin-styled-components/pull/117#discussion_r161072446).

kitten:
@existentialism I never understood why we haven't set that up in the first place. If someone here is up for PRing this that'd be AWESOME :tada:

damassi:
Closing in favor of #134

PR #129, opened 2018-02-28
disable "uglifyPure" mode by default

### LINKED ISSUE #126: Upgrading from 1.4.0 to 1.5.0 makes webpack builds significantly slower
After upgrading from 1.4.0 to 1.5.0 our webpack build time increased from ~30sec to ~130sec. We're using styled-components 2.4.0 and I've also checked with styled-components 3.1.0. Any hints appreciated.

probablyup:
unfortunately this feature seems to dramatically inflate the
amount of time babel takes to process files... perhaps there is
an unintended loop happening or something of the like

fixes #126


kitten:
Let's get rid of this :tada:

PR #132, opened 2018-03-19
New Option: cssTemplate

LordofMaps:
I am part of team that has created a react-based website that is embeddable on other websites. We use and **love** styled-components however we frequently have our styles inadvertently overridden by the sites that host our content. The main issue is the specificity of the class selectors added by styled-components combined with our reluctance to use injectGlobal — it would most-likely override the styles on the site embedding our content.

Does that make sense? 

The most straightforward way to fix this is to wrap all of our styles in somethings like `&& { /* our style here */ }`. The issue with this approach has been enforcement as the size of our code-base grows.

So I figured I would write a babel plugin that did it for us and _hopefully_  do it in such a way that others would be able to use it as well.

One thing I know is missing from this PR is formal documentation. I will gladly add that once I know where it should go.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ import uglifyPure from './visitors/uglifyPure'
 import minify from './visitors/minify'
 import displayNameAndId from './visitors/displayNameAndId'
 import templateLiterals from './visitors/templateLiterals'
+import cssTemplate from './visitors/cssTemplate'
 import assignStyledRequired from './visitors/assignStyledRequired'
 import { noParserImportDeclaration, noParserRequireCallExpression } from './visitors/noParserImport'
 
@@ -20,6 +21,7 @@ export default function({ types: t }) {
         minify(path, state)
         displayNameAndId(path, state)
         templateLiterals(path, state)
+        cssTemplate(path, state)
       },
       VariableDeclarator(path, state) {
         assignStyledRequired(path, state)
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -71,3 +71,8 @@ export const isHelper = (tag, state) => (
   isCSSHelper(tag, state) ||
   isKeyframesHelper(tag, state)
 )
+
+export const isExtend = (tag, state) => (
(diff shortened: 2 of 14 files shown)


LordofMaps:
One more thing. I would like a few good opinions on the detector for "extend". I hope there is a better/more specific way to could be done.  I also seem to remember that there is conversation about removing extend completely but until then I need it to be parsed,

LordofMaps:
@mxstbr thanks for the quick reply. An external plugin makes sense. There was just so much in common with this plugin that starting with a branch was the quickest route.
Also, I cannot believe I missed that plugin. 🙄   Although I don't think it would completely solve my use case.

kitten:
@LordofMaps yes, sorry about that. The plugin wasn't released for a while and we lost track of it, so it's not on the site yet. But maybe they'd appreciate it, if you'd take a look and maybe even contribute some of your code back to their plugin?

Also, while I'm saying that, if you happen to reuse some of the code here for their plugin, don't forget to include the licence that comes with this repo — bla bla legal annoyance bla.

@mxstbr I think it'd be very reasonable to include this plugin and link to it from this section of the docs too: https://www.styled-components.com/docs/advanced#existing-css

LordofMaps:
@kitten I've created a separate plugin which is also under MIT. Thanks for the feedback.

kitten:
@LordofMaps super! I took a quick look; make sure to add it to the awesome-styled-components list as soon as it’s old enough (we have a minimum time that must pass after open sourcing your repo to ensure maturity)

I also took a quick look st the licence file since I touched on that ;) it won’t be enough to use the same licence since it uses code that is copyrighted under our licence. You’ll either need to link to our licence for each file that originated here, or—which is way easier—add the copyright notice to your licence file I.e. add the “Copyright 20XX by ...” next to yours.

Sorry, obviously not a lawyer, but since I’ve noticed this I’d just like to make sure that we cover all grounds 👍

Thanks for working on this btw, hopefully we can add it to our docs 🙌

PR #137, opened 2018-05-08
Look for displayName value in class properties

msokk:
We're using styled-components as static class properties like this:
```js
class Modal {
  static Title = styled.h2``;
  static Body = styled.article``;
}

const Container = styled.div`
  ${Modal.Title} {
    color: #4a4a4a;
  }
`;
```

I updated `getName` function to look for class property too - either regular property or static one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "license": "MIT",
   "devDependencies": {
     "babel-cli": "^6.24.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-preset-env": "^1.2.2",
     "jest": "^17.0.3",
     "prettier": "^1.12.1",
--- a/src/utils/getName.js
+++ b/src/utils/getName.js
@@ -18,6 +18,9 @@ export default path => {
       // const X = { Y: styled }
     } else if (path.isObjectProperty()) {
       namedNode = path.node.key
+      // class Y { (static) X = styled }
+    } else if (path.isClassProperty()) {
+      namedNode = path.node.key
       // let X; X = styled
     } else if (path.isVariableDeclarator()) {
       namedNode = path.node.id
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
   "license": "MIT",
   "devDependencies": {
     "babel-cli": "^6.24.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
```
evan-scott-zocdoc: doesn't this need to be a hard dep?

kitten:
While we're at it, are we using object properties too? e.g.

```
const Anim = {
  Text: styled.div``
}
```

package.json
msokk: @evan-scott-zocdoc I double checked to be sure, but yeah this is only needed for test to work.

Method `path.isClassProperty()` works without this dependency too. In our use case we are doing the babel transpilation for styled-components before Typescript with `@babel/plugin-syntax-typescript` + some more (see https://github.com/gaearon/react-hot-loader#add-babel-before-typescript)

msokk:
@kitten It already works with objects - https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/utils/getName.js#L19
output being `<filename>__Text` in your case.

kitten:
@msokk Perfect! So we'll just need a little rebase :)

msokk:
@kitten rebased.

mxstbr:
Thank you so much for helping us improve styled-components! Based on our [Community Guidelines](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md) every person that has a PR of any kind merged is offered **an invitation to the styled-components organization**—that is you right now!

Should you accept, you'll get write access to the main repository. (and a fancy `styled-components` logo on your GitHub profile!) You'll be able to label and close issues, merge pull requests, create new branches, etc. We want you to help us out with those things, so don't be scared to do them! Make sure to read our [contribution and community guidelines](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md), which explains all of this in more detail. Of course, if you have any questions just let me know!

PR #139, opened 2018-05-09
correctly prefix component displayName starting with a number

### LINKED ISSUE #138: Invalid class name being generated (first character is number) causing styles to fail
It appears babel-plugin-styled-components is assigning component classnames based on folder/file path structure. For example I have a component

```
/14-offer-radbtn/index.js
```

that file contains a component

```
const HoverIndicator = styled.span`
    opacity: 0;
`;
const StyledLink = const StyledLink = styled.a`
    &:hover {
        ${HoverIndicator} {
            opacity: 1;
        }
    }
`;

const BA14OfferRadbtn = () => (
    <StyledLink>
        <HoverIndicator />
    </StyledLink>
);

export default BA14OfferRadbtn;
```

However the hover effect doesn't work. On closer inspection, it appears that the generated class name is invalid as it starts with a number and is being ignored by the browser
image

generated CSS rule: 
image

It appears that the class name is generated by looking at the folder path of the file. As the containing folder begins with "14" the class name follows suit invalidly. 

## System:
 - OS: macOS High Sierra 10.13.4
 - CPU: x64 Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
 - Memory: 361.34 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 8.9.4 - ~/.nvm/versions/node/v8.9.4/bin/node
 - npm: 5.8.0 - ~/.nvm/versions/node/v8.9.4/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.5.0 => 1.5.1
 - styled-components: ^3.2.3 => 3.2.6

## Expected Behavior
Valid CSS class name generated. 

## Actual Behavior
Invalid CSS classname generated, starting with a number, which is ignored by the browser

probablyup:
closes #138, when the displayName is derived from the directory name, there's a chance the directory could start with a number. This ends up being an invalid CSS class name as they must begin with a hyphen or ASCII letter.

split up the commits so the noise from adding prettier is isolated


probablyup:
@mxstbr done 👍

PR #150, opened 2018-08-12
rewrite styled import to "no tags" entry

probablyup:
~~Depends on https://github.com/styled-components/styled-components/pull/1912~~ merged


src/visitors/rewriteStyledImport.js
```diff
@@ -0,0 +1,14 @@
...
+import { isStyled } from '../utils/detectors'
+
+/**
+ * When using the Babel plugin, we desugar styled.div to styled('div'), which means we can
+ * then use a lighter-weight version of s-c since those element names don't need to be kept around
+ * ahead of time.
+ */
+export default path => {
+  if (path.node.source.value === 'styled-components') {
+    path.node.source = t.stringLiteral('styled-components/no-tags')
```
kitten: This means though that we’re currently breaking styled-components/native and /primitives if we’d merge this and this starts being on by default?

probablyup: I think as of right now it would only break `styled-components/native` due to the implicit resolution from https://github.com/styled-components/styled-components/blob/f401eda35ef16065509ea8a7375290fb8b5858a6/package.json#L9

Maybe we should remove that and just have people do this again

```js
import styled from 'styled-components/native'
```

probablyup: Seems safer overall and less prone to bundler wonkiness

kitten: @probablyup yea, sorry, the comment was half-baked anyway.

I’m not sure whether we should be going back on that change or not. Currently I’m even thinking tempting people to use universal switches by suggesting it ourselves is bad.

The reality is React Native and Web are _extremely_ different and shared styles should be explicit and mindful. So we should probably deprecate it again. The alternative would be to add a no-tags.native.js to the index to fix it up automatically, which I think is a neat compromise for now.

Long term the RN library should maybe even be separate from styles-components tbh. API-parity is important, sure, but I’d rather have RN specific changes and a playground there and port stuff manually.

cc @mxstbr

probablyup: > no-tags.native.js

Yeah I can definitely do that. I'm 50/50 on what to do about native in general. Keeping it in the same repo definitely helps keep the API in sync, but it's true that how it works is so very different.

kitten:
@probablyup @mxstbr yea, I think it just checks whether the target is a string. I don’t think any constructWithOptions path has knowledge of whether it was called from a shorthand.

src/visitors/rewriteStyledImport.js
mxstbr: Yeah honestly we should probably revert that change :confused: It's just made typing impossible and I don't think we should encourage that usage...

PR #153, opened 2018-08-21
DX improvements & cleanup

probablyup:


probablyup:
I was thinking we should even remove the ability to turn "ssr" mode off. Not sure when stable hashes would ever be an undesired thing.

PR #154, opened 2018-08-21
preparations for v4 bump

probablyup:


README.md
```diff
@@ -1,13 +1,18 @@
...
+* better debugging through automatic annotation of your styled components based on their context in the file system, etc.
+* various types of minification for styles and the tagged template literals styled-components uses
 
 ## Quick start
 
 Install the plugin first:
 
 ```
-npm install --save-dev babel-plugin-styled-components
+npm i -D babel-plugin-styled-components
```
kachkaev: Shorthands like this one may look a bit too magical for beginners. What is the drawback of keeping the long notation?

probablyup: No drawback, just more verbose

kachkaev: That’s good for commands in the docs, isn’t it? :)

mxstbr: Yeah I think it should go back to the original version to be clearer for beginners, it doesn't hurt :blush:

PR #163, opened 2018-09-11
reintroduce "pure" tagging

probablyup:
I found and fixed some issues with the original implementation which is why it probably was so slow.

```
npm i --save-dev babel-plugin-styled-components@experimental
```

`.babelrc`

```json
{
  "plugins": [
    ["babel-plugin-styled-components", { "pure": true }]
  ]
}
```


probablyup:
Can test it with `babel-plugin-styled-components@experimental`

probablyup:
The only thing I'm not sure about is technically `keyframes` is not pure prior to v4, but if you don't interpolate it somewhere then you're not using the animation... so it does it matter?

probablyup:
Yeah it's off by default for now

brad-decker:
Thanks for your work here! <3

PR #171, opened 2018-10-18
use closures with preserved names as option handler functions

manpenaloza:
this pr aims to simplify this file in terms of:
- no need to actively declare new functions `state` gets passed to
- no need to pass through `state` down the runtime chain

### DIFF
--- a/src/utils/options.js
+++ b/src/utils/options.js
@@ -1,14 +1,12 @@
-function getOption({ opts }, name, defaultValue = true) {
+const getOption = (name, defaultValue = true) => ({ opts }) => {
   return opts[name] === undefined || opts[name] === null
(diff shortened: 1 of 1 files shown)


quantizor (approved):
Makes sense

probablyup:
Needs a rebase

quantizor:
Thanks @manpenaloza, and apologies this sat for so long.

`src/utils/options.js` was rewritten during the 2.x line and the file already lives on the closure/factory shape you were proposing here. Each option is exported as a small `useX(state)` function that closes over `getOption`, so we no longer pass `state` through a runtime chain. The intent of this PR is effectively in.

Closing as superseded, with thanks for the original direction.

PR #172, opened 2018-10-18
[RFC] Add support for the css prop

mxstbr:
This integrates `babel-plugin-css-prop` by @satya164 (shoutout to him for being a genius) into our own Babel plugin, thus making the `css` prop work for all existing users!

<details>
  <summary>Click here to read more about how this works under the hood</summary>

The way this works is incredible, requires 0 changes to core. This is the crux:

```JS
const StaticString = props => <p css="flex: 1;">A</p>;

const CustomComp = props => <Comp css="flex: 1;">A</Comp>

// is turned into 👇👇👇👇👇

const _CSSP = styled.p`flex: 1;`;

const StaticString = props => <_CSSP>A</_CSSP>

const _CSSP1 = styled(Comp)`flex: 1;`;

const CustomComp = props => <_CSSP1>A</_CSSP1>;
```

I've also moved the entire test suite into this repo, check that out to see some of the edge cases (inc interpolations, dynamic interpolations, global interpolations,...) this covers

</details>

<br />

The big question is: do we want to have native support for the `css` prop?

/cc the core team (@kitten @probablyup @imbhargav5) and interested onlookers (@andrewingram @jxnblk).

----

Todos before shipping:

- [ ] Write docs

Potential future optimisations, probably not necessary for v1:

- [ ] Optimize object style usage (`<p css={{ color: 'blue' }} />`) to not use a dynamic prop
- [x] Optimize css helper usage (`<p css={css\`color: blue;\`} />`) to not use a dynamic prop


mxstbr:
I'd personally vote for yes, for the following reasons:

1. It's very useful in the early stages of a project, when it's not entirely clear where the boundaries of components should go. Our current API is amazing once you know where the abstraction should be, but doesn't allow you to move as quick as the `css` prop. To quote @andrewingram from [here](https://spectrum.chat/thread/47206c1b-a688-424e-9e96-6f265993587e?m=MTUzOTc3OTI0NzA1OQ==): "I find that inline methods are better when you haven't settled on the abstractions yet, it's common for me to start with large mega components and clean things up as I understand the problems (and therefore appropriate abstractions) better."
2. Lots of people use `styled-system` and/or `rebass` already to hack around our missing support for it, but the downside of the user-land approach is that you have to use the custom component everywhere to be able to use the prop. This allows you to put it on any JSX node
3. There is a large demand from users for this API, as other libraries have shown—I've often heard it's a big reason people choose another library over s-c. (see [here](https://mobile.twitter.com/sebastienlorber/status/1043806926432358400), [here](https://mobile.twitter.com/kentcdodds/status/1043282998521561088) and [here](https://mobile.twitter.com/ryanflorence/status/907320595859124224))

What do you think?

AndrewIngram:
This might come across weirdly, but is there something that styled-components offers that libraries like Emotion don't? From an outside perspective, mindshare seems to be the biggest advantage, is there anything else?

mxstbr:
> From an outside perspective, mindshare seems to be the biggest advantage, is there anything else?

See my thoughts here: https://www.reddit.com/r/reactjs/comments/9d583z/announcing_styledcomponents_v4_better_faster/e5fud05/

mxstbr:
> Doesn't seem to support arrays, eg css={[margin-left: 1px;,margin-right: 1px]), this is mainly useful for composing style primitives together.

I'm not sure about this, it would essentially transpile down to

```JS
const Bla = styled.div`
  ${['margin-left: 1px']}
`
```

which I think would work?

jxnblk:
Personally, I really like having the `css` prop as an escape hatch to handle one-off style changes with a more React-like API; it helps abstract the CSS-in-JS library away as an implementation detail. I also like how Styled Components and Emotion have an *almost* 1:1 API, and adding this to reduce the differences between the two libraries seems like an obvious move.

That said, I'm not really a fan of syntax being too tightly coupled with build tools. It's similar to why I don't recommend using things like CSS Modules: because the syntax you write only works when you're using webpack and the CSS Modules loader. This makes the code a little less portable. For the `css` prop with Babel, it's a little different, but any JSX that relies on the `css` prop will need to have this plugin to work properly, whereas this can be done without any extra build configuration using regular React components.

But, that's just my two cents, and I generally think this would be a great addition 👍

johnnysprinkles:
This is great. Only thing it doesn't cover is the scenario where you both a) don't want to create an extra component and b) you want to apply the same CSS to multiple repeated elements. That's a scenario that emotion handles wells, with `css` tagged template literal declared outside the JSX and used as many places as you need.

That's ok though, can't have everything! Might be possible to add later if there's enough demand.

probablyup:
withComponent hasn’t been removed yet, or you can use the new “as” prop

mxstbr:
@johnnysprinkles This doesn't work?

```JS
const bla = css``;

<div css={bla}>
<span css={bla}>
```

If not we should defo make that work, /cc @satya164

satya164:
@mxstbr it should work. let's add a test for it

johnnysprinkles:
Oh cool, I didn't see any mention of using it that way in the new docs at https://github.com/styled-components/styled-components-website/pull/365/commits/9b6008aa9bd9d7f722a6b4233f5bc9812f40101e It would make sense that the css prop is just Babel sugar for what you describe above. Would the import in that case be `import { css } from styled-components`?

johnnysprinkles:
Oh cool, I didn't see any mention of using it that way in the new docs at styled-components/styled-components-website@9b6008a Would the import in that case be import { css } from styled-components? And the css prop can take either a string as documented or whatever type the `css` tagged template literal returns?

mxstbr:
Yep @johnnysprinkles!

mxstbr:
This doesn't correctly add the styled import. Even code as simple as this throws a "styled is undefined" error:

```JS
// @flow
import React from 'react';
import Card from '../../shared/components/Card';
import config from '../../../config';

export default () => (
  <div
    css={`
      width: 35em;
    `}
  >
    <Card>
      <h1>Login or Sign Up</h1>
      <p>
        <a href={config.API_URI + '/auth/google'}>
          Sign up or login with Google
        </a>
      </p>
    </Card>
  </div>
);
```

satya164:
I'll take a look asap

mxstbr:
Just hit another bug:

```
TypeError: Cannot read property 'replace' of undefined
  at /node_modules/babel-plugin-styled-components/lib/visitors/transpileCssProp.js:36:75
```

With this code:

```JS
// @flow
import React from 'react';
import Card from '../../shared/components/Card';
import Link from '../../shared/components/Link';
import { Button } from '../../shared/components/Button';
import config from '../../../config';

export default () => (
  <div
    css={`
      width: 100%;
      height: 100%;
      display: flex;
      align-items: center;
      justify-content: center;
    `}
  >
    <div
      css={`
        max-width: 35em;
        width: 100%;
        margin: 0 auto;
      `}
    >
      <Card>
        <h1 css="margin-top: 0;">Sign Up or Log In</h1>
        <Button.Ghost
          css="display: block;"
          href={config.API_URI + '/auth/google'}
        >
          Authenticate with Google
        </Button.Ghost>
        <p css="margin-bottom: 0;">
          Your information is private and secure. By signing up, you agree to the{' '}
          <Link href="/terms">Terms of Service</Link> and{' '}
          <Link href="/privacy">Privacy Policy</Link>.
        </p>
      </Card>
    </div>
  </div>
);
```

kopax:
Hi and thanks for the amazing work you've done so far.

May i know if the `css` props has a specification somewhere ? I'd like to learn about it.

If this is a new API for adding css to components, it will reduce the amount of tuning we need to do and reduce the amount of template literal which is good and code will look nicer. 

May I know how this will handle the `props`  (generally used for adding logic to the css computation) and the `withTheme` (user for adding logic to the js/css using theme values.

Thanks in advance,  this feature is very exciting, is there a target version yet?

mxstbr:
> May i know if the css props has a specification somewhere ? I'd like to learn about it.
> If this is a new API for adding css to components, it will reduce the amount of tuning we need to do and reduce the amount of template literal which is good and code will look nicer.

The specification is: put whatever you want in the `css` prop and we'll turn it into a styled component :wink:

> May I know how this will handle the props (generally used for adding logic to the css computation) and the withTheme (user for adding logic to the js/css using theme values.

See the test suite in this PR for more in-depth examples, but it handles props and variables like any other styled component, with the added benefit of static interpolations:

```JSX
const Bla = (props) => (
  <div css={css`
    border-radius: ${props.borderRadius};
    color: ${p => p.color};
    background: ${p => p.theme.background};
  `}
   color="blue"
 />
)

<Bla borderRadius="3px" />
```

I hope that makes sense!

No target release date yet, but you can test this already if you `npm install --save babel-plugin-styled-components@experimental `! Note that it still has some rough edges, as my most recent comments illustrated.

satya164:
Thanks for finding the bugs @mxstbr. I'll take look over the weekend!

mxstbr:
@satya164 last two commits _should_ fix those issues, mind giving them a double check? :pray:

mxstbr:
Note: Should change the prefix of the generated display name to "Styled" rather than "CSS".

kopax:
@mxstbr thanks a lot for the insight, we'll test it soon. I believe we can't use it yet for distributing a module right? Does this require a plugin to be working or will styled be able to handle css props by itself?

mxstbr:
> . I believe we can't use it yet for distributing a module right? 

You totally can, it just might be a bit buggy! `npm install --save-dev babel-plugin-styled-components@experimental`

> What is the plugin exactly for? Does this require a plugin to be working or will styled be able to handle css props by itself?

It only requires the plugin at build-time, the users of the distributed module do not need to use the plugin! Good question, we should probably add that to the docs

src/visitors/transpileCssProp.js
```diff
@@ -19,14 +19,10 @@ export default t => (path, state) => {
...
       state.items.push(
-        t.variableDeclaration('var', [
-          t.variableDeclarator(
-            t.identifier('styled'),
-            t.callExpression(t.identifier('require'), [
-              t.stringLiteral('styled-components'),
-            ])
-          ),
-        ])
+        t.importDeclaration(
```
satya164: Looks good. Though maybe just doing `const styled = require('styled-components').default` will be better than adding an import since it seems to add a lot of extra code.

src/visitors/transpileCssProp.js
```diff
@@ -2,6 +2,14 @@
 // @see https://github.com/satya164/babel-plugin-css-prop
 import { useCssProp } from '../utils/options'
 
+const getTag = node => {
+  if (typeof node.name === 'string') return node.name
+  if (node.type === 'JSXMemberExpression') {
```
satya164: I usually prefer to use it like `t.isJSXMemberExpression(node)` instead of checking type manually

src/visitors/transpileCssProp.js
```diff
@@ -0,0 +1,104 @@
+// Most of this code was taken from @satya164's babel-plugin-css-prop
+// @see https://github.com/satya164/babel-plugin-css-prop
+import { useCssProp } from '../utils/options'
+
+const getTag = node => {
+  if (typeof node.name === 'string') return node.name
+  if (node.type === 'JSXMemberExpression') {
```
satya164: I usually prefer to use it like `t.isJSXMemberExpression(node)` instead of checking type manually

src/visitors/transpileCssProp.js
```diff
@@ -0,0 +1,104 @@
...
+    return `${getTag(node.object)}.${node.property.name}`
+  }
+  return undefined
```
satya164: Maybe throw a syntax error here instead of returning undefined so it's easier to identify where the error occurred? Something like:

```suggestion
  throw path.buildCodeFrameError('Failed to get the name');
```

src/visitors/transpileCssProp.js
mxstbr: I think going with imports make more sense, since some people might not even transpile imports potentially?

src/visitors/transpileCssProp.js
mxstbr: Great call

mxstbr:
Going to ship this as per internal discussions!

src/visitors/transpileCssProp.js
```diff
@@ -0,0 +1,108 @@
+// Most of this code was taken from @satya164's babel-plugin-css-prop
+// @see https://github.com/satya164/babel-plugin-css-prop
+import { useCssProp } from '../utils/options'
+
+const getName = (node, t) => {
+  if (typeof node.name === 'string') return node.name
+  if (t.isJSXMemberExpression(node)) {
+    return `${getName(node.object, t)}.${node.property.name}`
+  }
+  throw path.buildCodeFrameError(
```
satya164: Oops! This function needs to accept `path` as an argument (or be defined inside the plugin closure to access path)

mxstbr: Can you submit a PR? Thanks! :pray: (and let's hope nobody hits this anytime soon... :wink:)

satya164: Will try to do it soon!

PR #183, opened 2018-11-27
Switch from @babel/plugin-syntax-jsx to babel-plugin-syntax-jsx to restore Babel v6 compat

mxstbr:


mxstbr:
Trying this out locally to fix https://github.com/A-gambit/CSS-IN-JS-Benchmarks/pull/44

mxstbr:
Works! We should ship this asap, otherwise people who upgrade from 1.x to 1.9.0 are going to have a bad time 😅

Issue #186, opened 2018-11-30
`cssProp` feature produces invalid CSS with object syntax


amakk:
Hi, I'm using the new `css` property, and I noticed that when I upgraded to babel-plugin-styled-components `1.9.2` (from `1.9.0`) it is no longer handling objects in CSS property correctly.

Given this input:

```html
<div css={{ color: 'red' }}>Test</div>
```

I am getting this output: 
```css
.kIHpqD{-webkit-undefinedcolor:red;-ms-flex-undefinedcolor:red;undefinedcolor:red;-webkit-undefined;-ms-flex-undefined;undefined;}
```
and further down: 
```javascript
<div class="my-page___StyledDiv-sc-1f0w8lp-3 kIHpqD">Test</div>
```

I'm expecting it to output the css as something like: 
```css
.kIHpqD{color:red;}
```

If it helps, I'm using `styled-components 4.1.2` with Gatsby via `gatsby-plugin-styled-components 3.0.4`

mxstbr:
Hmm, this seems to be a bug, strange. @satya164 I think this was caused by the `requeue`?

satya164:
That's pretty weird. No idea why it would add `-webkit-undefined` etc. as prefix. It'll be great if you can create a repro so I can test.

mxstbr:
I can replicate this, not sure why this broke.

satya164:
Is there a repro somewhere that I can check?

satya164:
Can someone please create a repo with the repro?

alexlouden:
I'm seeing the same issues after moving from v1.8 with gatsby & rebass:
```js
import { Box } from 'rebass'
<Box p={3} color="blue" css={{ border: "1px solid red" }} {...props} />
```

with babel-plugin-styled-components@1.9.0 & 1.9.1 during `gatsby build`:
```js
  Error: component---src-templates-docstemplate-jsx-d6ef5c80a52750f5e099.js from UglifyJs
  SyntaxError: Unexpected character '`' [component---src-templates-docstemplate-jsx-d6ef5c80a52750f5e099.js:5956,62]
```

and with 1.9.2:
```js
  Module build failed: TypeError: <filepath>.js: Property value expected type of string but got null
```

Basic repro here - doesn't work online but works if you download & run `yarn && yarn start` locally - https://codesandbox.io/s/4z11joq1z7
[...]

PR #191, opened 2018-12-10
Add better support for objects in the css prop

mxstbr:
This patch makes it so the Babel plugin turns objects in the css prop into components directly, rather than passing the object via a prop:

```JS
<div css={{ color: 'blue' }} />

// is turned into
const StyledDiv = styled.div({
  color: 'blue'
});

<StyledDiv />
```

This might help with https://github.com/styled-components/babel-plugin-styled-components/issues/186, have to test and verify though.

/cc @satya164 for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.9.2",
+  "version": "1.9.3-0",
   "name": "babel-plugin-styled-components",
   "description": "Improve the debugging experience and add server-side rendering support to styled-components",
   "repository": "styled-components/babel-plugin-styled-components",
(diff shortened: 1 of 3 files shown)


satya164:
The reason I passed it as a prop is because the object can refer to variables in the scope. Maybe the error is because of a missing semicolon after the interpolation?

mxstbr:
Ohh damn good catch.

I don't think the semi-colon should matter, but let me test it

mxstbr:
I wonder what else could be causing this. It has to be our custom tagged template literal transformation, right?

satya164:
Yup. I can take a look later if you can create a repo with the setup :)

PR #198, opened 2018-12-18
Use babel test for testing the babel plugin

satya164:
babel-test supports writing transformed code to individual output.js files while still integrating with Jest snapshots. This means that we get syntax highlighting in the transformed files making it easier to spot mistakes.


satya164:
cc @mxstbr

PR #199, opened 2018-12-20
chore: remove sc peerDependencies

evenchange4:
Remove npm warning if I want to create a custom babel preset as a **npm library** with `babel-plugin-styled-components`:

```bash
warning " > babel-plugin-styled-components@1.10.0" has unmet peer dependency "styled-components@>= 2".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,9 +37,6 @@
   "resolutions": {
     "babel-core": "7.0.0-bridge.0"
   },
-  "peerDependencies": {
-    "styled-components": ">= 2"
-  },
   "scripts": {
     "clean": "rimraf lib",
     "style": "prettier --write src/**/*.js",


stevemao:
Hey @evenchange4 , thanks for the PR. I'm wondering if you put `styled-components@>= 2` as the peer dep (and probably dev dep) of your own custom babel preset, will you still see the warning?

stevemao:
> Put styled-components@>= 2 as the peer dep still see the warning.

I feel like npm should suppress the warning because you are propagating the peer dep to your end user.

stevemao:
> Maybe I should install sc as the dev dep (but actually I do not use sc in the preset library).

I'm not sure how you test your preset without sc. Maybe using it to test is the right way

mxstbr:
Yeah I don't think it makes sense to remove this, given that the Babel plugin only works in conjunction with styled-components. It seems like having the peerDep there is the right choice for us, so I'll close this for now.

PR #202, opened 2018-12-26
Pass dynamic components in the as prop

satya164:
I haven't tried this in a real app :)

### DIFF
--- a/test/fixtures/transpile-css-prop/code.js
+++ b/test/fixtures/transpile-css-prop/code.js
@@ -32,8 +32,6 @@ const CssHelperProp = p => (
  * Dynamic prop
  */
 
-let Paragraph;
-
 const CustomComp = p => <Paragraph css="flex: 1">H</Paragraph>
 
 const DynamicProp = p => <p css={props.cssText}>H</p>
@@ -88,8 +86,6 @@ const DynamicCssHelperProp = p => (
     A
   </p>
 )
-
-let Button;
 
 const CustomCompWithDot = p => <Button.Ghost css="flex: 1">H</Button.Ghost>
 
--- a/test/fixtures/transpile-css-prop/code.js
+++ b/test/fixtures/transpile-css-prop/code.js
@@ -32,6 +32,8 @@ const CssHelperProp = p => (
  * Dynamic prop
  */
 
+let Paragraph;
+
 const CustomComp = p => <Paragraph css="flex: 1">H</Paragraph>
 
 const DynamicProp = p => <p css={props.cssText}>H</p>
@@ -87,8 +89,26 @@ const DynamicCssHelperProp = p => (
   </p>
 )
 
+let Button;
+
 const CustomCompWithDot = p => <Button.Ghost css="flex: 1">H</Button.Ghost>
 
 const NestedCompWithDot = p => (
   <Button.Ghost.New css="flex: 1">H</Button.Ghost.New>
 )
+
+const DynamicComp = p => {
+  const { Box } = p.components;
+
+  <Box css="color: blue" title="Hello">X</Box>
+}
+
+const DynamicCompWithDot = p => {
+  <p.components.Box css="color: blue" title="Hello">Y</p.components.Box>
+}
+
+const DynamicCompWithAs = p => {
+  const { Box, Pyramid } = p.components;
+
+  <Box css="color: blue" as={Pyramid}>Z</Box>
+}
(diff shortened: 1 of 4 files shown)


satya164:
This won't work if the variable is re-assigned. Maybe we should always use the as prop?

test/fixtures/transpile-css-prop/output.js
```diff
@@ -293,6 +339,12 @@ var _StyledP11 = (0, _styledComponents.default)("p")(_templateObject15(), functi
   return props.theme.color;
 });
 
-var _StyledButtonGhost = (0, _styledComponents.default)(Button.Ghost)(_templateObject16());
+var _StyledButtonGhost = (0, _styledComponents.default)(undefined)(_templateObject16());
+
+var _StyledButtonGhostNew = (0, _styledComponents.default)(undefined)(_templateObject17());
+
+var _StyledBox = (0, _styledComponents.default)(undefined)(_templateObject18());
```
mxstbr: I'm pretty sure this will throw an error? I don't think we support this dynamic-as-prop-only usage.

test/fixtures/transpile-css-prop/code.js
```diff
@@ -92,3 +92,19 @@ const CustomCompWithDot = p => <Button.Ghost css="flex: 1">H</Button.Ghost>
...
 )
+
+const DynamicComp = p => {
+  const { Box } = p.components;
+
+  <Box css="color: blue" title="Hello">X</Box>
+}
+
+const DynamicCompWithDot = p => {
+  <p.components.Box css="color: blue" title="Hello">Y</p.components.Box>
```
mxstbr: Isn't this invalid JSX since all custom components need to be upper case? Maybe change it to

```JSX
P => <P.components.Box ... />
```

stevensacks:
Please let me know when I can pull and test this PR in my app and I'll let you know how it works! 👍

test/fixtures/transpile-css-prop/output.js
satya164: @mxstbr yeah, you're right. what do you suggest? maybe some random string? it shouldn't really matter since the component to render will always be passed in the `as` prop.

test/fixtures/transpile-css-prop/code.js
satya164: @mxstbr it's valid. custom components need to start with uppercase only if it's a variable name, for property access etc, it doesn't matter. example: https://codesandbox.io/s/ql0mzp2jl6

satya164:
@stevensacks you can give it a try now and let us know

stevensacks:
Will do. Happy new year!

stevensacks:
@satya164 I've been trying to test this pull request but I'm getting an error:

```json
"babel-plugin-styled-components": "git://github.com/styled-components/babel-plugin-styled-components.git#fd610c1e083f92678e8133fe807192d0d7cfba7c",

**Comment by stevensacks** [comment]
I tried that and I got the following error.

```bash
Error: Cannot find module 'styled-components/macro' from '/Users/steven/Development/project/src/components/App'
```
error
 
Both of the libs are showing up as not installed.

satya164:
I've no idea why this will happen. Maybe @mxstbr has encountered it?

Worst case, you could clone the branch, run `npm build` and copy it to your `node_modules` for trying.

stevensacks:
@mxstbr Would you be willing to merge this into a release candidate to make things easier?

mxstbr:
I'll publish this under a tag—one second...

mxstbr:
```
yarn add --dev babel-plugin-styled-components@experimental
```

stevensacks:
Thanks! I’ll give it a shot.

stevensacks:
I'm hoping I have the correct version.

```json
{
  "babel-plugin-styled-components": "^1.10.1-0",
}
```

It doesn't work, though.

@satya164 here's how you can test it yourself in a simple way. The below code throws the error.

```javascript
import {css, keyframes} from 'styled-components';
import styled from 'styled-components/macro';
import React from 'react';

const UI = {
  View: ({className, children}) => (
    <div className={className}>{children}</div>
  ),
};

const App = () => {
  const {View} = UI;
  return (
    <View css={viewStyle}>
      <p>Testing 1 2 3</p>
    </View>
  );
}

// From create-react-app's default css
const viewStyle = `
    background-color: #282c34;
    min-height: 100vh;
    display: flex;
    flex-direction: column;
    align-items: center;
    justify-content: center;
    font-size: calc(10px + 2vmin);
    color: white;
`;
```

Let me know how that works!

satya164:
@stevemao sorry, I didn't have much time to investigate. will try this weekend.

stevensacks:
@satya164 Anything I can do to make working on this easier?

PR #219, opened 2019-04-17
feat: read environment specific configuration

codepunkt:
It would be great if babel-plugin-styled-components could read environment specific option when using via `babel-plugin-macros` instead of directly using this plugin.

As far as i know, `babel-plugin-macros` supports configuration via `cosmiconfig` from package.json or various .rc or .config files, but it doesn't allow environment specifics.

I'd love to use styled-components via macros without configuring this plugin manually, but still be able to do things like `{ displayName: true }` in development and `{ displayName: false }` in production.

This PR enables that usecase.


codepunkt:
No comment from reviewers, alternate solution possible via `babel-config-macro` javascript configuration file. Guess this isn't wanted, so i closed it.

PR #225, opened 2019-06-15
fix css prop transpilation for stable server-side rendering

probablyup:


src/index.js
```diff
@@ -10,8 +10,15 @@ export default function({ types: t }) {
   return {
     inherits: syntax,
     visitor: {
-      JSXAttribute(path, state) {
-        transpileCssProp(t)(path, state)
+      Program(path, state) {
```
probablyup: All visitors are run simultaneously, so we need to create our new styled components in a preliminary pass over the code, _then_ run all the other visitors.

PR #227, opened 2019-06-22
fix ssr with styled HOC + object styles

probablyup:


src/visitors/displayNameAndId.js
```diff
@@ -137,9 +137,18 @@ export default t => (path, state) => {
...
       ? isStyled(t)(path.node.tag, state)
-      : isStyled(t)(path.node.callee, state) &&
-        path.node.callee.property &&
-        path.node.callee.property.name !== 'withConfig'
+      : /* styled()`` */ (isStyled(t)(path.node.callee, state) &&
+          path.node.callee.property &&
+          path.node.callee.property.name !== 'withConfig') ||
+        // styled(x)({})
+        (isStyled(t)(path.node.callee, state) &&
+          !t.isMemberExpression(path.node.callee.callee)) ||
```
tm1000: Shouldn't this be checking for a property name of withConfig like the other check do?

Issue #203, opened 2019-01-09
cssProp isn't working with parcel


nathanspecht:
## Description

The cssProp feature for babel-plugin-styled-components does not work with Parcel.

## Error

```
Invariant Violation
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `App`.
```

## Dependencies

```
"babel-plugin-styled-components": "1.10.0"
"react": "16.4.2"
"react-dom": "16.4.2"
"styled-components": "4.1.3"
"@babel/cli": "^7.0.0"
"@babel/core": "^7.0.0"
"@babel/plugin-proposal-class-properties": "^7.0.0"
"@babel/preset-env": "^7.0.0"
"babel-core": "7.0.0-bridge.0"
"babel-plugin-transform-es2015-modules-commonjs": "^6.26.2"
"babel-minify": "^0.4.3"
"babel-plugin-transform-class-properties": "^6.24.1"
"babel-preset-es2015": "^6.24.1"
"babel-preset-react": "^6.24.1"
"babel-preset-stage-3": "^6.24.1"
"parcel-bundler": "^1.6.1"
```

## .babelrc

```
{
  "presets": [
    "@babel/preset-env",
    "@babel/preset-react"
  ],
  "plugins": [
    ["babel-plugin-styled-components", {
      "cssProp": true
    }],
    ["@babel/plugin-proposal-class-properties", { "loose": true }]
  ]
}
```

## CodeSandbox

https://codesandbox.io/s/j727ky0p55

LanhKhoc:
Btw, it doesn't work with **@babel/plugin-transform-react-constant-elements** too.

![image](https://user-images.githubusercontent.com/35893600/53547898-ce91b480-3b62-11e9-9b9f-b4d0501d6b74.png)

probablyup:
Hmm I think this is because of injection order... have some ideas on how to fix it.

probablyup:
Ok yeah I've got a fix for this in #228. Example sandbox using a test build: https://codesandbox.io/s/react-parcel-starter-o2wov?fontsize=14

probablyup:
@LanhKhoc please make a separate issue for that with a reproduction sandbox :)

probablyup:
Closed via #228 - it'll go out with the next patch release or you can try it now with: `babel-plugin-styled-components@ej-test`

PR #228, opened 2019-06-24
fix injection location of generated components via css prop

### LINKED ISSUE #221: css prop not working with babel plugins: "transform-react-inline-elements", "transform-react-constant-elements"
Hi,

I'm using [react-boilerplate](https://github.com/react-boilerplate/react-boilerplate) for a project and have upgraded styled-components to v4. 

For the production build I use the following babel configuration (i've only added `styled-components`, the rest is the initial configuration provided by the boilerplate):

```
"production": {
        "only": [
          "app"
        ],
        "plugins": [
          "styled-components",
          "transform-react-remove-prop-types",
          "transform-react-inline-elements",
          "transform-react-constant-elements"
        ]
      },
```

I have a component which uses the `css` prop, but even though it is working fine in development, when I run the production build the styles defined in the `css` prop are not being applied unless I remove `transform-react-inline-elements` and `transform-react-constant-elements` from the babel configurations.

I wasn't able to tell why it's happening, just reporting this here in case anyone has any clues. For now there's no workaround besides not using the `css` prop.

Thank you!

**Edit:** [Also confirmed here](https://github.com/styled-components/babel-plugin-styled-components/issues/203#issuecomment-468163379)

---

Example:

```
<div css={{ padding: '5rem', border: '1px solid red' }}>Dummy div</div>
```

`package.json`:

```
"babel-cli": "6.26.0",
    "babel-core": "6.26.3",
    "babel-eslint": "^8.2.6",
    "babel-loader": "7.1.4",
    "babel-plugin-dynamic-import-node": "1.2.0",
    "babel-plugin-react-intl": "2.4.0",
    "babel-plugin-react-transform": "3.0.0",
    "babel-plugin-styled-components": "^1.10.0",
    "babel-plugin-transform-es2015-modules-commonjs": "6.26.2",
    "babel-plugin-transform-react-constant-elements": "6.23.0",
    "babel-plugin-transform-react-inline-elements": "6.22.0",
    "babel-plugin-transform-react-jsx-source": "6.22.0",
    "babel-plugin-transform-react-remove-prop-types": "0.4.13",
    "babel-preset-env": "1.7.0",
    "babel-preset-react": "6.24.1",
    "babel-preset-stage-0": "6.24.1",
```

probablyup:
inject the generated components before the tree they're being used in rather than to the end of the file, this fixes situations where a bundler might wrap a component in an IIFE and our component generated from the css prop hasn't been initialized yet (fixes #221, #203)

### DIFF
--- a/src/visitors/transpileCssProp.js
+++ b/src/visitors/transpileCssProp.js
@@ -45,11 +45,20 @@ export default t => (path, state) => {
   )
 
   let styled
+  let injector
 
   if (TAG_NAME_REGEXP.test(name)) {
     styled = t.memberExpression(importName, t.identifier(name))
   } else {
     styled = t.callExpression(importName, [t.identifier(name)])
+
+    if (bindings[name] && !t.isImportDeclaration(bindings[name].path.parent)) {
+      injector = nodeToInsert =>
+        (t.isVariableDeclaration(bindings[name].path.parent)
+          ? bindings[name].path.parentPath
+          : bindings[name].path
+        ).insertAfter(nodeToInsert)
+    }
   }
 
   let css
@@ -179,9 +188,17 @@ export default t => (path, state) => {
     }, [])
   }
 
-  // Add the tagged template expression and then requeue the newly added node
-  // so Babel runs over it again
-  const length = program.node.body.push(
+  if (!injector) {
+    let parent = elem
+
+    while (!t.isProgram(parent.parentPath)) {
+      parent = parent.parentPath
+    }
+
+    injector = nodeToInsert => parent.insertBefore(nodeToInsert)
+  }
+
+  injector(
     t.variableDeclaration('var', [
       t.variableDeclarator(
         id,
@@ -191,6 +208,4 @@ export default t => (path, state) => {
       ),
     ])
   )
-
-  program.requeue(program.get('body')[length - 1])
 }
--- a/test/fixtures/transpile-css-prop-add-import/output.js
+++ b/test/fixtures/transpile-css-prop-add-import/output.js
@@ -27,6 +27,8 @@ function _templateObject() {
 
 function _taggedTemplateLiteral(strings, raw) { if (!raw) { raw = strings.slice(0); } return Object.freeze(Object.defineProperties(strings, { raw: { value: Object.freeze(raw) } })); }
 
+var _StyledDiv = _styledComponents["default"].div(_templateObject());
+
 var _default = function _default() {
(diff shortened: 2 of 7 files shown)


src/visitors/transpileCssProp.js
```diff
@@ -120,7 +120,13 @@ export default t => (path, state) => {
     }, [])
   }
 
-  program.node.body.push(
+  let parent = elem
+
+  while (!t.isProgram(parent.parentPath)) {
```
probablyup: I assume this should be a safe thing to do since before we were basically doing the same thing but an append instead of a sibling prepend to the node tree that we're currently working on.

probablyup:
This is available to test under `babel-styled-components@1.10.2-ej0` if anyone wants to.

probablyup:
Waiting for codesandbox to work again so I can test these changes against the babel macro

probablyup:
I'm going to extract out the object css prop and @bpierre's changes into a separate branch, I'm getting weird errors building projects after the insert position changes and it needs more bake time

probablyup:
Tested against a babel macro project and a fairly hefty nextjs project with `babel-plugin-styled-components@1.10.2-ej8` and it looks good.

probablyup:
cleaned up the PR - all the content is the same as before just compacted and rebased

UXDart:
I'm having this problem when using rollup to bundle, everything seems to be working but the generated code has the styledDiv creation after and outside the React.createElement, so I have error saying createElement received undefined. I assume this is because rollup ast must be different? what can I do? TIA!

UXDart:
it seems this is the code that sets the injector, if the injector is null, then it adds to the program body, but at the end... for some reason

`  if (TAG_NAME_REGEXP.test(name)) {
    styled = t.callExpression(importName, [t.stringLiteral(name)]);
  } else {
    styled = t.callExpression(importName, [nameExpression]);

    if (bindings[name] && !t.isImportDeclaration(bindings[name].path.parent)) {
      injector = nodeToInsert => (t.isVariableDeclaration(bindings[name].path.parent) ? bindings[name].path.parentPath : bindings[name].path).insertAfter(nodeToInsert);
    }
  }
`

how can I get into the else part of the if? not sure what name should be and where... thx

UXDart:
oh... nvm just found why this happens, I was using css prop on the main jsx file, I used it now inside a second file, and it works. it seems if css prop is used on main file, it is injecting the new var outside the main scope. thx

PR #229, opened 2019-06-25
fix object syntax css prop, support more types of components & custom elements

probablyup:
Reduced version of #228 since the injection location code isn't totally ready yet

1. fix transpilation of objects being passed into the css prop

2. modifications from @bpierre to support custom elements, lowercase component variable names with dots, nested component names with dots


probablyup:
Going to take a second pass at #228 tonight and see if I can get it working consistently before I cut another patch release. If I can't get it where it needs to be will do a patch with these changes tomorrow.

PR #230, opened 2019-06-27
do not try to transpile things we don't own

### LINKED ISSUE #224: Doesn't play well with material-ui v4
In the newest major update of Material-UI, there is a `styled` function that is *somewhat* similar to `styled-components`. When importing it like:

```ts
import { styled } from "@material-ui/styles"
```

the following error is thrown:

```
[0] const CloseButton = styled(IconButton)({
[0] ^
[0] 
[0] TypeError: styles_1.styled(...).withConfig is not a function
[0] 
[0]   57 | };
[0]   58 | 
[0] > 59 | const CloseButton = styled(IconButton)({
[0]   60 |   position: "absolute",
[0]   61 |   top: 10,
[0]   62 |   right: 10,
[0] 
[0]     at Object../src/common/SomeComponent (/.../SomeComponent.tsx:59:1)
```

However when the imported variable's name is changed, it works. example:

```ts
import { styled as matStyled } from "@material-ui/styles"
```
As best as I can tell this is happening because this babel plugin is seeing the `styled` import and transforming it.

probablyup:
Fixes #224


src/utils/detectors.js
```diff
@@ -17,7 +17,11 @@ export const importLocalName = (name, state, bypassCache = false) => {
     return localNameCache[cacheKey]
   }
 
-  let localName = name === 'default' ? 'styled' : name
+  let localName = state.styledRequired
+    ? name === 'default'
+      ? 'styled'
+      : name
+    : false
```
probablyup: For the ESM case we want to start with the assumption there is no import and work from there, rather than assume the `styled` import is present and named in that precise way. In the case of the `css` prop we then will dynamically-inject the import as needed.

Issue #232, opened 2019-06-28
Next.js app stuck on "compiling" with version >= 1.10.3


znarf:
Our Next.js app is https://github.com/opencollective/opencollective-frontend/

When upgrading to a version >= `1.10.3` of `babel-plugin-styled-components`, our dev environment is unusable, the app is stuck on "compiling ...".

We had to downgrade to version `1.10.2`:
https://github.com/opencollective/opencollective-frontend/pull/2044

**To reproduce:**
- `git clone git@github.com:opencollective/opencollective-frontend.git`
- `cd opencollective-frontend`
- `npm install`
- `npm install babel-plugin-styled-components@1.10.3 --save-exact`
- `npm run dev`
- Go to: `http://localhost:3000/`

**Current behavior**:
- app is stuck on "compiling ..."
- can't load any page.

**Expected behavior**:
- app succeeds compiling
- can load any page

probablyup:
can you try 1.10.4?

znarf:
Same for `1.10.4`.

We actually first encountered this behavior in `1.10.4`, we later tried `1.10.3` with the same result. So, this should have been introduced in `1.10.3`.

c0b41:
same as gatsby, build never completed..
![Ekran Alıntısı](https://user-images.githubusercontent.com/2834954/60388975-37158180-9ac2-11e9-8102-e245bf4a24d5.PNG)

nghiepit:
Same issue
`1.0.2` work fine.

Swampy821:
Adding on this, this broke our build within Docker and took us a long time to pinpoint where it was coming from. You may want to consider reverting back to 1.10.2 to save people that have it marked as `^1.10.0` like we did while you figure out the issue.

probablyup:
Hard to figure out the issue when no one has provided a reproduction.

znarf:
@probablyup I think I gave precise steps to reproduce. Let me know if you need anything else from me.

Vanuan:
Hello from Gatsby users:
https://github.com/gatsbyjs/gatsby/issues/15205

Vanuan:
I think you can use this example: https://github.com/gatsbyjs/gatsby/tree/master/examples/using-styled-components

Vanuan:
People say disabling source maps fixes the issue. Hope it helps

znarf:
[...]

PR #233, opened 2019-07-01
Fix `importName` type confusion (string, bool, node)

ZauberNerd:
Apparently `importName` can have many different types:
https://github.com/styled-components/babel-plugin-styled-components/issues/232#issuecomment-507223968

This commit ensures that the check whether the name is in the scope
(bindings) does not break on plain strings.

Closes: #232

### DIFF
--- a/src/visitors/transpileCssProp.js
+++ b/src/visitors/transpileCssProp.js
@@ -29,7 +29,7 @@ export default t => (path, state) => {
   const { bindings } = program.scope
 
   // Insert import if it doesn't exist yet
-  if (!importName || !bindings[importName.name]) {
+  if (!importName || !bindings[importName.name] || !bindings[importName]) {
     addDefault(path, 'styled-components', {
       nameHint: 'styled',
     })


agriffis:
@ZauberNerd I know this is a couple years later, but are you still around to talk about this?

I'm confused by this line of code:

```
   if (!importName || !bindings[importName.name] || !bindings[importName]) {
```

It seems to me that this will _always_ evaluate to true. Either `importName` is a string, so `bindings[importName.name]` will be undefined, or `importName` is an object, so `bindings[importName]` will be undefined.

Should it be:

```
   if (!importName || !bindings[importName.name || importName]) {
```

or maybe a `typeof` check?

ZauberNerd:
@agriffis looks like you're correct and it was an oversight on my part.

agriffis:
@ZauberNerd Thanks for the reply! Not sure yet about next steps on this, I think any change in this area would deserve additional tests to make sure it's working as intended. But I appreciate your following up to my question.

PR #234, opened 2019-07-01
Fix early return for cached `false` values in `importLocalName`

ZauberNerd:
Closes: #232

### DIFF
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -13,7 +13,7 @@ const localNameCache = {}
 export const importLocalName = (name, state, bypassCache = false) => {
   const cacheKey = name + state.file.opts.filename
 
-  if (!bypassCache && localNameCache[cacheKey]) {
+  if (!bypassCache && cacheKey in localNameCache) {
     return localNameCache[cacheKey]
   }
 


src/utils/detectors.js
```diff
@@ -13,7 +13,7 @@ const localNameCache = {}
 export const importLocalName = (name, state, bypassCache = false) => {
   const cacheKey = name + state.file.opts.filename
 
-  if (!bypassCache && localNameCache[cacheKey]) {
+  if (!bypassCache && cacheKey in localNameCache) {
```
ZauberNerd: Before this change the value in the cache would be `false` and therefore this `if` condition would never evaluate truthy and we would traverse all `import`s for files that are not part of styled-components every time again.

probablyup: Hmm I'm confused when this would ever happen... is it possible for the same babel plugin to be run in parallel on the same file? The only place where the cached value could be `false` is in the css prop transpile visitor, and it immediately injects and re-caches the injected import...

probablyup:
Published a test build under `babel-plugin-styled-components@test` if you want to share it and see if it works for people.

c0b41:
@probablyup working great for gatsby probably work next js too..

@ZauberNerd great work btw 👍👍👍

probablyup:
@c0b41 to be clear nextjs has always been working for me. I have 5 nextjs projects using s-c and they never had this issue, so I have a strong suspicion there's a mixed babel environment problem happening that manifests more strongly with certain project setups.

src/utils/detectors.js
ZauberNerd: @probablyup as per your comment (https://github.com/styled-components/babel-plugin-styled-components/pull/230#discussion_r298018008) it is false for all ESM files: https://github.com/styled-components/babel-plugin-styled-components/blob/2bfc83a6fb350165db7a926788065693069cd025/src/utils/detectors.js#L20-L24 because as far as I can tell `styledRequired` will only be set for commonJS files (https://github.com/styled-components/babel-plugin-styled-components/blob/e04eb6f264f89a8abe5f32b0ca22dc200fe7e360/src/visitors/assignStyledRequired.js).
And then `localName` is initially set to `false` (https://github.com/styled-components/babel-plugin-styled-components/blob/2bfc83a6fb350165db7a926788065693069cd025/src/utils/detectors.js#L24) and if the file has no `styled-components` import (https://github.com/styled-components/babel-plugin-styled-components/blob/2bfc83a6fb350165db7a926788065693069cd025/src/utils/detectors.js#L31) it will never re-define the `localName` variable, therefore the cache value of `localNameCache[cacheKey]` will be `false` (https://github.com/styled-components/babel-plugin-styled-components/blob/2bfc83a6fb350165db7a926788065693069cd025/src/utils/detectors.js#L53) which means that this `if` condition won't evaluate truthy (https://github.com/styled-components/babel-plugin-styled-components/blob/v1.10.5/src/utils/detectors.js#L16).

probablyup: All that is true but the only time `localName` will ever stay false is in the use case of the css prop which doesn't require `styled` to be in scope (except the babel macro use case.) Otherwise `localName` has to be set to something picked up via `ImportDeclaration` traversal because you can't use the normal features of styled-components without importing it.

There is a code smell here...

probablyup: If you use yarn, could you do me the favor of reverting to 1.10.4 in your project and adding a "resolutions" field to your master package.json to force any installed versions of `babel-plugin-styled-components` to the same version? e.g.

```json
"resolutions: {
  "babel-plugin-styled-components": "1.10.4"
}
```

ZauberNerd: @probablyup I think you are right, but the problem is all other files that do not use styed-components (in which case the `localName` would stay `false`). For me it looks like this traversal (https://github.com/styled-components/babel-plugin-styled-components/blob/2bfc83a6fb350165db7a926788065693069cd025/src/utils/detectors.js#L26-L51) is very costly (that's probably why the cache was introduced in the first place).

ZauberNerd: @probablyup for my use-case you could try and `yarn create hops-app my-app` and then edit your `node_modules` folder with some `console.log`'s or the changes of this PR to see the effect. I think it might even work without these changes but it will take a huge amount of time longer and that is what is breaking the builds (because travis, for example, terminates builds that don't print outpout for longer than 10 minutes).

PR #237, opened 2019-07-09
Skip `css` function call when the template literal is pure

kevin940726:
## `skipPureString`

Consider the following snippet

```js
const flex = css`
  display: flex;
`;

const Container = styled.div`
  ${flex}
`;
```

Since the `flex` css is actually a pure template literal (i.e. no interpolations), we could omit the `css` function call to save us some bytes and prevent an unneeded function call. The below snippet is identical to the above.

```js
const flex = 'display: flex;';

const Container = styled.div`
  ${flex}
`;
```

But sometimes we would want to keep the `css` tagged template, like when we want IDE syntax highlighting support, or prettier support. Then it results in an additional `css` call.

Most of the time, it shouldn't really matters. The footprint is minimal and performance cost is also unnoticeable. But in the other way around, it also wouldn't harm to let babel plugin to skip the `css` tag for us, a free optimization.

I'm calling this feature `skipPureString`, and so is the feature flag in babel config options. The naming is totally TBD, as at the time I couldn't come up with a better name 😅.

Note that the order of this feature is placed immediately below the `minify` step, so that the content of the tag can still be minified.

This feature should be fully backward-compatible (unless someone decided to use undocumented feature or api, like type checking a `css` return value) since it skips all of the _unsure_ tagged templates with expressions in them. Because of that, the feature is turned on by default.

### TODOs, Caveats
- [x] Add tests
- [ ] Update documentation in [styled-components-website](https://github.com/styled-components/styled-components-website)
- [ ] Since we will skip the `css` tag by default, we could actually encourage the use of `css` whenever possible, for syntax highlighting, prettier, minifier support out of the box.
- [ ] The only caveat I can think of is that if someday we implement additional feature in `css` and makes skipping it somehow unfeasible. We could just remove this feature then?


test/fixtures/transpile-template-literals-with-config/code.js
```diff
@@ -5,7 +5,7 @@ const Named = styled.div`
 `
 
 const NamedWithInterpolation = styled.div`
-  color: ${color => props.color};
+  color: ${props => props.color};
```
kevin940726: I believe I've caught a typo 😂

kevin940726:
Close it until I found a better way to do this.

Issue #209, opened 2019-02-19
Ability to write full path to a component's file in className


rifler:
Hi there!

Our common pattern is to write template in `index.tsx` (or `view.tsx`) and styles in `styles.tsx`. So the most common component dir structure looks like:
client/components/AwesomeComponent:
 - index.tsx
 - styles.tsx

That's why classes looks like `styles__Header-sc-1ei6mxn-18` and you can't easily identify what `Header` is it, from `Component1/styles.tsx` or from `Component2/styles.tsx`. It would be great for us to write full path to a file from cwd in className, like so `client_components_component1_styles__Header-sc-1ei6mxn-18`.

What do you think?

btw we absolutely love styled components

probablyup:
What do you think @mxstbr? My only concern is this could greatly increase the payload size of both the CSS and the HTML, though some of it would get gzipped away. Perhaps it could make sense behind a flag that isn't enabled by default.

rifler:
+1 for flag, it should be used in development only

elie222:
I added a flag for it in this PR:
https://github.com/styled-components/babel-plugin-styled-components/pull/259

PR #259, opened 2019-11-11
Add option to use parent folder name as component name

elie222:
Addresses:
https://github.com/styled-components/babel-plugin-styled-components/issues/209

### DIFF
--- a/src/utils/options.js
+++ b/src/utils/options.js
@@ -7,6 +7,7 @@ function getOption({ opts }, name, defaultValue = true) {
 export const useDisplayName = state => getOption(state, 'displayName')
 export const useSSR = state => getOption(state, 'ssr', true)
 export const useFileName = state => getOption(state, 'fileName')
+export const useFolderName = state => getOption(state, 'folderName', false)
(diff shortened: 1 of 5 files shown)


src/utils/options.js
```diff
@@ -7,6 +7,7 @@ function getOption({ opts }, name, defaultValue = true) {
 export const useDisplayName = state => getOption(state, 'displayName')
 export const useSSR = state => getOption(state, 'ssr', true)
 export const useFileName = state => getOption(state, 'fileName')
+export const useFolderName = state => getOption(state, 'folderName', false)
```
probablyup, line 10: I think this could be named more obviously... perhaps `prefixComponentWithFolderName`

igorokb: `folderName` sounds not less clearer as `fileName`. Context matters.

quantizor:
Thanks @elie222. The folder-name option (enhancement #209) hasn't moved into the default behavior but the option-add approach here is sound. Leaving open as the candidate implementation for #209 when we get to it.

PR #266, opened 2020-01-24
emit styled('tag') instead of styled.tag for css prop

probablyup:
this handles if the given tag is not in the dom elements whitelist yet in styled-components proper

fixes https://github.com/styled-components/styled-components/pull/2965#issuecomment-574537642

### DIFF
--- a/src/visitors/transpileCssProp.js
+++ b/src/visitors/transpileCssProp.js
@@ -49,7 +49,7 @@ export default t => (path, state) => {
   let injector
 
   if (TAG_NAME_REGEXP.test(name)) {
-    styled = t.memberExpression(importName, t.identifier(name))
+    styled = t.callExpression(importName, [t.stringLiteral(name)])
   } else {
     styled = t.callExpression(importName, [t.identifier(name)])
 
--- a/test/fixtures/transpile-css-prop-add-import/output.js
+++ b/test/fixtures/transpile-css-prop-add-import/output.js
@@ -27,7 +27,7 @@ function _templateObject() {
 
 function _taggedTemplateLiteral(strings, raw) { if (!raw) { raw = strings.slice(0); } return Object.freeze(Object.defineProperties(strings, { raw: { value: Object.freeze(raw) } })); }
 
-var _StyledDiv = _styledComponents["default"].div(_templateObject());
+var _StyledDiv = (0, _styledComponents["default"])("div")(_templateObject());
 
 var _default = function _default() {
   return <_StyledDiv>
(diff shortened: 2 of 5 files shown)


test/fixtures/transpile-css-prop-add-import/output.js
```diff
@@ -27,7 +27,7 @@ function _templateObject() {
 
 function _taggedTemplateLiteral(strings, raw) { if (!raw) { raw = strings.slice(0); } return Object.freeze(Object.defineProperties(strings, { raw: { value: Object.freeze(raw) } })); }
 
-var _StyledDiv = _styledComponents["default"].div(_templateObject());
+var _StyledDiv = (0, _styledComponents["default"])("div")(_templateObject());
```
probablyup: I'm not sure why Babel is adding this sequence (`(0, fn)`), seems unnecessary.

probablyup: Apparently it's a strict-mode thing: https://stackoverflow.com/questions/32275135/why-does-babel-rewrite-imported-function-call-to-0-fn

src/visitors/transpileCssProp.js
```diff
@@ -49,7 +49,7 @@ export default t => (path, state) => {
   let injector
 
   if (TAG_NAME_REGEXP.test(name)) {
```
kitten: This check seems fishy to me. Wouldn't we want to identify from the AST whether the element name is a JSX identifier or element name? Or does the AST not offer any hints on checking that?

probablyup: There isn't a clear understanding of if it's a scope variable or not, for example:

```
 Node {
        type: 'JSXIdentifier',
        start: 147,
        end: 150,
        loc:
         SourceLocation {
           start: Position { line: 7, column: 3 },
           end: Position { line: 7, column: 6 } },
        name: 'div' }
```

```
Node {
        type: 'JSXIdentifier',
        start: 1518,
        end: 1524,
        loc:
         SourceLocation {
           start: Position { line: 104, column: 34 },
           end: Position { line: 104, column: 40 } },
        name: 'Thing3' }
```

If you go and try to infer from local scope bindings there are also false positives:

```js
 console.log(
    path.scope.hasBinding(getVariableName(elem.node.name, t)),
    getVariableName(elem.node.name, t)
  )
```

```
false 'Button'
    console.log src/visitors/transpileCssProp.js:61
      false 'Button.Ghost'
    console.log src/visitors/transpileCssProp.js:61
      false 'button'
    console.log src/visitors/transpileCssProp.js:61
      false 'button-ghost'
    console.log src/visitors/transpileCssProp.js:61
      true 'p'
```

There's some variable called "p" in scope which conflicts with the tag named "p" and would trigger `styled(p)` which is wrong in this case.

probablyup:
Going to merge this for the moment - I don't think we should touch the existing tag checking code since it seems to be working fine

Issue #71, opened 2017-07-01
styled-components/native support


ckknight:
As-is, this plugin does not appear to detect usages of `styled` when imported from `styled-components/native` instead of `styled-components`.

patrickkempff:
+1 for displayName debugging

AndrewMorsillo:
This would be awesome! I just tried to set babel-plugin-styled-components up in my react-native project only to discover that displayName didn't work unfortunately.

Jekiwijaya:
This is needed for react-native-web 0.12. Since `styled-components` give className through the component, and on `react-native-web@0.12.x` className is ignored.

axhamre:
Is this still not working?

PR #267, opened 2020-02-02
feat: add options isNative to support default import styled-component…

Jekiwijaya:
add options isNative to support default import styled-components/native

Fixed: #71 

```
FEAT: add `isNative` option to make default import become styled-components/native
```

### DIFF
--- a/src/utils/options.js
+++ b/src/utils/options.js
@@ -14,3 +14,4 @@ export const useTranspileTemplateLiterals = state =>
 export const usePureAnnotation = state => getOption(state, 'pure', false)
 
 export const useCssProp = state => getOption(state, 'cssProp', true)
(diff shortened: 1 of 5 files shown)


Jekiwijaya:
@probablyup 
Currently default import for `css` props is `required('styled-components')`.

The behavior of `styled-components` and `styled-components/native` is different
https://github.com/necolas/react-native-web/issues/1146#issuecomment-476674280

In `react-native-web@0.12.x` the babel plugin css for styled-component will have no impact since it's wrapped the component using `required('styled-components').default` which override the className props and className is not supported in `react-native-web@0.12`.

To fix this, the babel plugin must be able to set the default import to `styled-components/native`

Thank you

patrickjm:
@Jekiwijaya Are you open to a PR adding support for `styled-components/primitives`?

PR #269, opened 2020-02-03
prefix all componentId with `sc-`

robilars-amazon:
Previously componentId would only be prefixed with `sc-` when component hash began with a digit. This differs from styled-components componentId generation where all componentIds are prefixed with `sc-`

This patch causes all  componentId to be prefixed with `sc-`


vxcamiloxv:
This fix works fine to me

PR #270, opened 2020-02-18
Added test for transpiling css prop as an array

taneba:
It seems possible to do things like:

```
style1 = css``
style2 = css``

<div css={[style1, style2]} />
```

But the test for this use case is missing now.


quantizor:
Thanks @taneba, and apologies for the long silence here.

Array-form `css={[mixin1, mixin2]}` is handled on current main: the array is passed through as a prop interpolation, same as a single mixin would be. The fixture diff in this PR predates the 2.x rewrite of the css-prop transform, so it can't be merged as-is, but the underlying behavior is exercised by the existing `transpile-css-prop` fixture coverage.

Closing as no-longer-applicable, with thanks for the original investigation.

PR #276, opened 2020-03-16
prefix all componentId with `sc-`

robilars-amazon:
Previously componentId would only be prefixed with `sc-` when component hash began with a digit. This differs from styled-components componentId generation where all componentIds are prefixed with `sc-`

This patch causes all componentId to be prefixed with `sc-`

Additionally, I have updated the build dependancies and rebuilt yarn.lock

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.10.7",
+  "version": "1.10.9",
   "name": "babel-plugin-styled-components",
   "description": "Improve the debugging experience and add server-side rendering support to styled-components",
   "repository": "styled-components/babel-plugin-styled-components",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -15,18 +15,18 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@babel/cli": "^7.4.4",
-    "@babel/core": "^7.4.5",
+    "@babel/cli": "^7.8.4",
+    "@babel/core": "^7.8.7",
     "@babel/plugin-proposal-class-properties": "^7.4.4",
-    "@babel/preset-env": "^7.4.5",
+    "@babel/preset-env": "^7.8.7",
```
probablyup, line 21: please drop unrelated changes, thanks

tobilen:
i believe this can be closed, since there is the `namespace' option that parameterizes this now

kitten:
I believe this is obsolete now given: https://github.com/styled-components/babel-plugin-styled-components/pull/281

PR #278, opened 2020-04-05
Add a failing test for minifying nested orphaned pseudo

Andarist:
The whitespace in `& :hover` has a semantic value - this is not equal to `&:hover` and thus this whitespace should not be removed when minifying quasis.

The problem is that [`compressSymbols`](https://github.com/styled-components/babel-plugin-styled-components/blob/55b43ee70bdd8cd20ba942577e3747879cf1b44b/src/minify/index.js#L58) is just too simplistic - parsing a language like CSS with RegExp is IMHO not the best idea in the long run as it loses the context in which particular symbols have been used and it's hard to develop further.

I've prepared a fix for this problem in Emotion: https://github.com/emotion-js/emotion/pull/1836 , we maybe could think of sharing the solution for this as the job is pretty much the same for both libraries. I'm afraid though that I won't have much time to abstract this away and port the solution here, but if you would like to do it I'm open for collaboration.


agriffis:
Oh cool, this should be merged after #376.

quantizor:
Thanks @Andarist. This is still on the radar. The minifier whitespace handling is one of the bigger remaining surface areas and the Emotion parser-based approach you pointed at is the obvious target. Not pulling it into the current bug-fix bundle to keep the diff scoped; will revisit as its own PR.

PR #279, opened 2020-04-06
Fix displayName broken by nested assignment

rozzzly:
Hi. This PR resolves an issue I ran into while using this plugin. Namely, when the plugin attempts to pull the identifier out of an `AssignmentExpression` which is nested (eg `A = B = 'foobar'`) it will choose the innermost identifier (`B`) while the others are ignored. This (as I explain below) may not be desirable. 

This PR adds a `preferOuterAssignment` option for your babel config which enables this alternate behavior.

#### Quick example:
For the following snippet:
```js
export const DoubleAssignment = Discarded = styled.div`
  width: 50%;
`;
```
if your babel config has:
```json
  "plugins": [
    ["babel-plugin-styled-components", {
		"ssr": false,
		"fileName": false,
		"displayName": true,
		"transpileTemplateLiterals": false
	}]
  ]
```
your emit will look like:
```js
export const DoubleAssignment = Discarded = styled.div.withConfig({
  displayName: "Discarded"
})`width:50%;`;
```
by enabling the `preferOuterAssignment` you'll get an emit which looks like:
```js
export const DoubleAssignment = Discarded = styled.div.withConfig({
  displayName: "DoubleAssignment"
})`width:50%;`;
```
### Real-world example where this be necessary
I first noticed this when some of my components were not getting their `displayName` set correctly while others where.

relevant snippet:
```jsx
import React from 'react';
import ContentContainer from './ContentContainer';

export const ScaffoldingContainer = styled.div`
	background: red;
`;

export const ScaffoldingWrapper = styled.div`
    margin: 4px;
`;

export const Scaffolding = ({ children }) => (
    <ScaffoldingContainer>
        <ScaffoldingWrapper>
            <ContentContainer>
                { children }
            </ContentContainer>
        </ScaffoldingWrapper>
    </ScaffoldingContainer>
);

export default Scaffolding;

```

After some experimentation, I found that by moving `ScaffoldingContainer` and `ScaffoldingWrapper` into a seperate file, things got a bit better:

_**`fileName` is `false`.** Therefor where the components are located should not determine their `displayName`_

I dug deeper and discovered the issue when looking at the emitted code.
```js
var ScaffoldingContainer = _c = styled_components__WEBPACK_IMPORTED_MODULE_2__["default"].div.withConfig({
  displayName: "_c",
  componentId: "sc-19b0fwj-0"
})(_templateObject());
var ScaffoldingWrapper = _c2 = styled_components__WEBPACK_IMPORTED_MODULE_2__["default"].div.withConfig({
  displayName: "_c2",
  componentId: "sc-19b0fwj-1"
})(_templateObject2());
var Scaffolding = function Scaffolding(_ref) {
  var children = _ref.children;
  return /*#__PURE__*/react__WEBPACK_IMPORTED_MODULE_1___default.a.createElement(ScaffoldingContainer, null, /*#__PURE__*/react__WEBPACK_IMPORTED_MODULE_1___default.a.createElement(ScaffoldingWrapper, null, /*#__PURE__*/react__WEBPACK_IMPORTED_MODULE_1___default.a.createElement(_ContentContainer__WEBPACK_IMPORTED_MODULE_3__["default"], null, children)));
};
_c3 = Scaffolding;
/* harmony default export */ __webpack_exports__["default"] = (Scaffolding);

var _c, _c2, _c3;

$RefreshReg$(_c, "ScaffoldingContainer");
$RefreshReg$(_c2, "ScaffoldingWrapper");
$RefreshReg$(_c3, "Scaffolding");
```

It's actually `react-refresh/babel` ([the successor to `react-hot-loader`](https://github.com/facebook/react/issues/16604#issuecomment-528663101)) that is causing the issue. It looks like when components in the same file reference each other it injects some helper code and creates this issue with nested assignment (ie: `var ScaffoldingContainer = _c = ...`). When `ScaffoldingContainer` and `ScaffoldingWrapper` were moved to a seperate file and they did reference other components (unlike `Scaffolding`) that code never got injected.

### So why am I PR'ing this repo and not `facebook/react`?

It's quite simple. The likelihood that if I reported this in `facebook/react` (and possibly [`pmmmwh/react-refresh-webpack-plugin`](https://github.com/pmmmwh/react-refresh-webpack-plugin)) have the react team decide it's a large enough issue that they rework their experimental mechanism and so anytime in the near future is a lot less likely than a simple PR here which is opt-in (via `preferOuterAssignment`) despite being non-breaking (at least in this repo's test suite). This has the added benefit of _(possibly)_ resolving other related issues.

### DIFF
--- a/src/utils/getName.js
+++ b/src/utils/getName.js
@@ -1,14 +1,16 @@
+import { preferOuterAssignment } from './options';
 /**
  * Get the name of variable that contains node
  *
  * @param  {Path} path to the node
+ * @param  {Object} state holds opts which might be of interest
  *
  * @return {String}   The target
  */
 
-export default t => path => {
+export default t => (path, state) => {
+  const prefersOuter = preferOuterAssignment(state)
   let namedNode
-
   path.find(path => {
     // const X = styled
     if (path.isAssignmentExpression()) {
@@ -28,7 +30,19 @@ export default t => path => {
     }
 
     // we've got an displayName (if we need it) no need to continue
-    if (namedNode) return true
+    // UNLESS this is an AssignmentExpression and we prefer to use
+    // the outermost VariableDeclarator; in that case we keep going
+    // because the parent of this path might be another AssignmentExpression
+    //
+    // example:
+    //   A = B = styled.div`color: red;`
+    // here we want A, not B (which is the current path) so we continue up
+    if (
+        namedNode &&
+        (!path.isAssignmentExpression() || !prefersOuter)
+    ) {
+      return true
+    }
   })
 
   // foo.bar -> bar
--- a/src/utils/options.js
+++ b/src/utils/options.js
@@ -14,3 +14,5 @@ export const useTranspileTemplateLiterals = state =>
 export const usePureAnnotation = state => getOption(state, 'pure', false)
 
 export const useCssProp = state => getOption(state, 'cssProp', true)
+export const preferOuterAssignment = state =>
+  getOption(state, 'preferOuterAssignment', false)
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -1,6 +1,9 @@
 import path from 'path'
 import fs from 'fs'
-import { useFileName, useDisplayName, useSSR } from '../utils/options'
+import {
+  useFileName, useDisplayName,
+  useSSR, preferOuterAssignment
+} from '../utils/options'
 import getName from '../utils/getName'
 import prefixLeadingDigit from '../utils/prefixDigit'
 import hash from '../utils/hash'
@@ -59,8 +62,8 @@ const getBlockName = file => {
 }
 
 const getDisplayName = t => (path, state) => {
-  const { file } = state
-  const componentName = getName(t)(path)
+  const file = useFileName(state) ? state.file : false;
+  const componentName = getName(t)(path, state)
   if (file) {
     const blockName = getBlockName(file)
     if (blockName === componentName) {
@@ -152,7 +155,7 @@ export default t => (path, state) => {
   ) {
     const displayName =
       useDisplayName(state) &&
-      getDisplayName(t)(path, useFileName(state) && state)
+      getDisplayName(t)(path, state)
 
     addConfig(t)(
       path,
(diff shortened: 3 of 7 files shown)


probablyup:
This needs a rebase

rozzzly:
@probablyup I'll rebase off master tonight

rozzzly:
Done.

probablyup:
@rozzzly what if we enabled this by default for NODE_ENV='development' since the refresh plugin isn't used in production or testing?

rozzzly:
@probablyup I added this as an opt-in flag because while it didn't break any of the tests and I can't think of any situations where this would break the `displayName` for anyone, didn't want to make that assumption.

If you're comfortable with that, you'll not hear any objections from me. It would probably simplify things as far as documentation goes: A few moments ago I realized, my previous commits never updated the `README.md` with the name of the option/explanation. When I checked the `README.md`, I found that the options are not documented there but on the [styled-components website](https://styled-components.com/docs/tooling#usage). The site has multiple sections on tooling all concatenated into one page, and I feel adding this long technical explanation multi-paragraph explanation about what `preferOuterAssigment` is and when/why it should be enabled would just be more clutter. I think it would be preferable to just have it be the default and briefer explanation without comparing emit code, etc.
> ## preferOuterAssigment
> Enabled by default since v1.11.2, this option ensures that when there is nested assignment such as:
> ```js
> export const DoubleAssignment = Discarded = styled.div`
>   width: 50%;
> `;
> ```
> that the component shows up as `DoubleAssignment` instead of `Discarded`. Even when your code does not contain nested assignment, this issue can be caused by other babel plugins changing the emitted code before it reaches `babel-plugin-styled-components`. For more details, see [PR#279](https://github.com/styled-components/babel-plugin-styled-components/pull/279)
>
> You can disable this feature by setting the `preferOuterAssignment` option to false:
> ```json
> {
>   "plugins": [
>     [
>       "babel-plugin-styled-components",
>       {
>         "preferOuterAssigment": false
>       }
>     ]
>   ]
> }

rockwotj:
Ha I ran into this myself and opened #355. Is there anyway I can help push this along to get merged and released?

rockwotj:
Also in terms of option or not. It seems that react-refresh always prefixes their names with an underscore so we could also only skip the inner assignment if it's an underscore name.

PR #283, opened 2020-04-28
Ensure cache miss to fix ESM compiling

nickhudkins:
In this discussion: https://github.com/styled-components/babel-plugin-styled-components/pull/234 that introduced the change to `cacheKey in localNameCache`, the author referenced a comment you made here:
https://github.com/styled-components/babel-plugin-styled-components/pull/230#discussion_r298018008

in which, you indicated that the `false` value there was to ensure that the traversal would continue again until you either picked up the import or dynamically injected it (I think?)

When the code was changed to account for this it failed to maintain the same behavior in the case that localName was intentionally left `false`.

This change ensures that the previous behavior (allowing the intentionally set `false` to fall through the cache lookup) remains, while ensuring a `false` value in the cache does not ALSO fall through.

Pardon the lack of supporting test. The only way I was able to reproduce this issue was when I was using rollup / babel to compile. I am still investigating what this difference is specifically to hopefully introduce a failing test prior to this commit.

### DIFF
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -4,6 +4,7 @@ const VALID_TOP_LEVEL_IMPORT_PATHS = [
   'styled-components/native',
   'styled-components/primitives',
 ]
+const CACHE_MISS = Symbol('CACHE_MISS');
 
 export const isValidTopLevelImport = x =>
   VALID_TOP_LEVEL_IMPORT_PATHS.includes(x)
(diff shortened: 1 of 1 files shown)


nickhudkins:
@probablyup welp, I am fully at a loss for how to test this. Do you have any guidance? I am also happy to cowboy it and say "It works with rollup, thanks"

marvinhagemeister:
Can confirm that this fixes the issue for me 👍

Issue #235, opened 2019-07-02
Spread object into `cssProp`  doesn't work


Grimones:
Hello there. Accidentally encountered an issue with spreading an object into css prop
Like in [polished](https://polished.js.org/docs/#fluidrange)

```jsx
<div
  css={{
    backgroundColor: 'palevioletred',
    ...fluidRange({
      prop: 'font-size',
      fromSize: '20px',
      toSize: '100px',
    }),
  }}
>
  hi
</div>
```

This error ocurred

```
TypeError: Property expression of JSXExpressionContainer expected node to be of a type ["Expression","JSXEmptyExpression"] but instead got null
    at Object.validate (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\@babel\core\node_modules\@babel\types\lib\definitions\utils.js:128:13)
    at validate (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\@babel\core\node_modules\@babel\types\lib\validators\validate.js:17:9)
    at builder (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\@babel\core\node_modules\@babel\types\lib\builders\builder.js:46:27)
    at Object.JSXExpressionContainer (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\@babel\core\node_modules\@babel\types\lib\builders\generated\index.js:767:31)
    at propertiesReducer (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\babel-plugin-styled-components\lib\visitors\transpileCssProp.js:124:83)
    at Array.reduce (<anonymous>)
    at C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\babel-plugin-styled-components\lib\visitors\transpileCssProp.js:110:39
    at PluginPass.JSXAttribute (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\babel-plugin-styled-components\lib\index.js:32:46)
    at NodePath._call (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\@babel\core\node_modules\@babel\traverse\lib\path\context.js:53:20)
    at NodePath.call (C:\Users\Andrew\Documents\projects-dev\lv-online\node_modules\@babel\core\node_modules\@babel\traverse\lib\path\context.js:40:17)
```

This doesn't happen only on `1.10.0` version.
Also this doesn't happen if you put the `fluidRange` into prop without spreading
```jsx
<div
[...]

PR #286, opened 2020-05-05
Support Spread Element

benjaminRomano:
**Background**
Spread operator does not work. See the following issue for more details: https://github.com/styled-components/babel-plugin-styled-components/issues/235.

This fails as the transpiler tries to convert the SpreadElement node into a key/value. Additionally, it needs to ensure that objects are converted into a p

**Change**
* Do not treat spreadElement as an expression
* Properly map object values of the SpreadElement argument into Styled component props.

**Test Plan**
Added a new test case that validates spread elements and nested spread elements are handled properly. The gen code looks correct, but I'm not too familiar with what styled components gen code should look like.

### DIFF
--- a/src/visitors/transpileCssProp.js
+++ b/src/visitors/transpileCssProp.js
@@ -127,7 +127,8 @@ export default t => (path, state) => {
 
         acc.push(property)
       } else if (
-        // if a non-primitive value we have to interpolate it
+        // if a spread element or a non-primitive value we have to interpolate it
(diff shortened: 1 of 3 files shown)


benjaminRomano:
@probablyup ptal when you get a chance

pie6k:
It would be great to have this merged.

kylemh:
@probablyup - friendly bump on this

PR #295, opened 2020-07-31
Support styled as a named export

will-stone:
To support the upcoming `topLevelImportPaths` option, this allows this babel plugin to pick-up `styled` as a named export:

```js
import { styled } from '@wrapping/package'`
```

config:

```
[
  'styled-components',
  {
    topLevelImportPaths: ['@wrapping/package'],
  },
]
```


probablyup (approved):
Thanks!

PR #297, opened 2020-08-17
move injected styled component to end of file

### LINKED ISSUE #260: Runtime error with css prop and const declared after component
Example:
```jsx
// in:
import React from "react"
import { css } from "styled-components"

export default function Example() {
  return <div css={someCss}>oops</div>
}

const someCss = css`
  color: red;
`

// out:
import _styled from "styled-components";
import React from "react";
import { css } from "styled-components";

var _StyledDiv = _styled.div.withConfig({
  displayName: "Example___StyledDiv",
  componentId: "sc-1j0h0n6-0"
})(["", ""], someCss);

export default function Example() {
  return <_StyledDiv>oops</_StyledDiv>;
}
const someCss = css(["color:red;"]);
```

At runtime, this code will fail, as it tries to access `someCss` before it's declared.

This works if we declare `someCss` before `Example`. The `someCss` style is correctly placed before the styled component.
```jsx
// in:
import React from "react"
import { css } from "styled-components"

const someCss = css`
  color: red;
`

export default function Example() {
  return <div css={someCss}>oops</div>
}

// out:
import _styled from "styled-components";
import React from "react";
import { css } from "styled-components";
const someCss = css(["color:red;"]);

var _StyledDiv = _styled.div.withConfig({
  displayName: "Example___StyledDiv",
  componentId: "sc-1j0h0n6-0"
})(["", ""], someCss);

export default function Example() {
  return <_StyledDiv>oops</_StyledDiv>;
}
```

Repro: https://github.com/kingdaro/styled-components-css-prop-error

Downgraded to 1.10.2 as a workaround for now. It looks like this commit introduced the change in behavior: https://github.com/styled-components/babel-plugin-styled-components/commit/9bc3c08ee31b82be1c2a74dc7f4a1fbd0e12ac7b

sfishel18:
in case it references a variable that is declared after the react component that uses it

fixes #260


probablyup (approved):
Seems ok to me

probablyup:
@sfishel18 can you fix this up against `main`?

PR #301, opened 2020-09-13
Recurse only when the argument of Spread Element is object expression

re-fort:
Using Spread syntax in css prop is currently supported, but this syntax is not supported.
```js
<div
  css={{
    ...obj,
    { color: 'red' }
  }}
>
  hi
</div>
```

it causes the error that can not read property `reduce` because it has not `property.argument.properties`

This PR resolves it by recursing only when the argument of Spread Element is object expression.


probablyup (approved):
LGTM, thanks!

Issue #268, opened 2020-02-03
jest-styled-components `toHaveStyleRule()` does not detect styles when using babel-plugin-styled-components


robilars-amazon:
### System:
 - OS: Linux 4.15 Debian GNU/Linux 9 (stretch) 9 (stretch)
 - CPU: (4) x64 Intel(R) Xeon(R) CPU @ 2.30GHz
 - Memory: 4.06 GB / 25.51 GB
 - Container: Yes
 - Shell: 4.4.12 - /bin/bash
### Binaries:
 - Node: 10.17.0 - /usr/local/bin/node
 - Yarn: 1.19.1 - /usr/local/bin/yarn
 - npm: 6.11.3 - /usr/local/bin/npm
### npmPackages:
 - babel-plugin-styled-components: ^1.10.7 => 1.10.7
 - jest-styled-components: ^7.0.0 => 7.0.0
 - styled-components: ^5.0.0 => 5.0.0

## Reproduction

https://repl.it/@robilarsamazon/jest-styled-components-5-and-babel-plugin-styled-components

## Steps to reproduce

Click "Run" to run execute unit test  

## Expected Behavior

Test passes

## Actual Behavior

Test fails 

## Notes

jest-styled-components `toHaveStyleRule()` depends on componentIds that start with `sc-`. When babel-plugin-styled-components is enabled, componentIds are not prefixed.

robilars-amazon:
The correct behavior can be seen by removing  `"babel-plugin-styled-components"` from babel.config.js and re-running the test in https://repl.it/@robilarsamazon/jest-styled-components-5-and-babel-plugin-styled-components

robilars-amazon:
Created PR #269 for a solution that makes all componentIds prefixed with `sc-`

vvanmol:
Hello, 

Thanks @robilars-amazon for the PR.

Any estimate on when this can be reviewed and merged ?

samplefrequency:
Love to see this get merged soon.

PatrickMcLennan:
Seconded, kinda driving me crazy over here.  Can try to help if needed.

robilars-amazon:
Looks like #290 was not building in CI. Created a new PR
https://github.com/styled-components/babel-plugin-styled-components/pull/275

robilars-amazon:
I published out a fork of the package with the PR applied. Feel free to use this till a fix is added to master
https://www.npmjs.com/package/@robilars/babel-plugin-styled-components

[...]

PR #313, opened 2021-02-11
Ensure sc- prefix is always added

chalbert:
Currently, the sc- is added at the start of the hash by replacing the first digit. The problem is that the hash may also start by a letter, causing the class name to sometimes miss the sc- prefix.

This fix makes it possible to use this plugin with styled-components@5 + jest-styled-components@7.

In the meantime, I've published the fix on npm under [@chalbert/babel-plugin-styled-components](https://www.npmjs.com/package/@chalbert/babel-plugin-styled-components)

In my experience, it fixes the following issues :

#268
styled-components/jest-styled-components#297
styled-components/jest-styled-components#335
styled-components/jest-styled-components#294
And potentially others.

### DIFF
--- a/test/fixtures/minify-single-line-comments-with-interpolations/output.js
+++ b/test/fixtures/minify-single-line-comments-with-interpolations/output.js
@@ -1,29 +1,29 @@
 import styled from 'styled-components';
 const Test1 = styled.div.withConfig({
   displayName: "code__Test1",
-  componentId: "kc0mjf-0"
+  componentId: "sc-kc0mjf-0"
 })(["width:100%;"]);
 const Test2 = styled.div.withConfig({
   displayName: "code__Test2",
-  componentId: "kc0mjf-1"
+  componentId: "sc-kc0mjf-1"
 })(["width:100%;"]);
 const Test3 = styled.div.withConfig({
   displayName: "code__Test3",
-  componentId: "kc0mjf-2"
+  componentId: "sc-kc0mjf-2"
 })(["width:100%;", ";"], 'red');
 const Test4 = styled.div.withConfig({
   displayName: "code__Test4",
-  componentId: "kc0mjf-3"
+  componentId: "sc-kc0mjf-3"
 })(["width:100%;"]);
 const Test5 = styled.div.withConfig({
   displayName: "code__Test5",
-  componentId: "kc0mjf-4"
+  componentId: "sc-kc0mjf-4"
 })(["width:100%;"]);
 const Test6 = styled.div.withConfig({
   displayName: "code__Test6",
-  componentId: "kc0mjf-5"
+  componentId: "sc-kc0mjf-5"
 })(["background:url(\"https://google.com\");width:100%;", " "], 'green');
 const Test7 = styled.div.withConfig({
   displayName: "code__Test7",
-  componentId: "kc0mjf-6"
+  componentId: "sc-kc0mjf-6"
 })(["background:url(\"https://google.com\");width:", ";", "  height:", ";"], p => p.props.width, 'green', p => p.props.height);
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -135,9 +135,7 @@ const getNextId = state => {
 
 const getComponentId = state => {
   // Prefix the identifier with a character because CSS classes cannot start with a number
-  return `${useNamespace(state)}${prefixLeadingDigit(
-    getFileHash(state)
-  )}-${getNextId(state)}`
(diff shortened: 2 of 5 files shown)


test/fixtures/minify-single-line-comments-with-interpolations/output.js
```diff
@@ -1,29 +1,29 @@
 import styled from 'styled-components';
 const Test1 = styled.div.withConfig({
-  displayName: "code__Test1",
-  componentId: "kc0mjf-0"
+  displayName: "sc-code__Test1",
```
probablyup, line 3: The `displayName` shouldn't be getting prefixed

chalbert: Right, I see. That may be why the bug was first introduced. By only prefixing digit, the goal was to skip prefixing the displayName, but it also skipped prefixing some of the hash. I'll look into it.

chalbert: @probablyup I've updated to prevent the prefixing of the displayName.

agriffis:
What about something like this instead?
- generated classes get consistently namespaced or `sc-` prefix
- block names get an underscore prefix when needed

```
diff --git a/src/utils/prefixDigit.js b/src/utils/prefixDigit.js
index 8cb8c4b..ec5d2ee 100644
--- a/src/utils/prefixDigit.js
+++ b/src/utils/prefixDigit.js
@@ -1,3 +1,3 @@
-export default function prefixLeadingDigit(str) {
-  return str.replace(/^(\d)/, 'sc-$1')
+export default function prefixLeadingDigit(str, prefix) {
+  return str.replace(/^(?=\d)/, prefix)
 }
diff --git a/src/visitors/displayNameAndId.js b/src/visitors/displayNameAndId.js
index 42c1c1f..eed1140 100644
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -72,8 +72,8 @@ const getDisplayName = t => (path, state) => {
       return componentName
     }
     return componentName
-      ? `${prefixLeadingDigit(blockName)}__${componentName}`
-      : prefixLeadingDigit(blockName)
+      ? `${prefixLeadingDigit(blockName, '_')}__${componentName}`
+      : prefixLeadingDigit(blockName, '_')
   } else {
     return componentName
   }
@@ -134,10 +134,8 @@ const getNextId = state => {
 }
 
 const getComponentId = state => {
-  // Prefix the identifier with a character because CSS classes cannot start with a number
-  return `${useNamespace(state)}${prefixLeadingDigit(
-    getFileHash(state)
-  )}-${getNextId(state)}`
+  // Ensure identifier has prefix to avoid invalid class starting with number.
+  return `${useNamespace(state) || 'sc-'}${getFileHash(state)}-${getNextId(state)}`
 }
 
 export default t => (path, state) => {
```

chalbert:
With this different approach, when using a namespace, the classname does not include `sc-` anymore. I don't know if there is a reason for that. It could be a breaking change for tools or projects that rely on having the `sc-` in the classname.

agriffis:
> With this different approach, when using a namespace, the classname does not include `sc-` anymore. I don't know if there is a reason for that. It could be a breaking change for tools or projects that rely on having the `sc-` in the classname.

It wasn't consistent previously, because of conditionalizing on the digit. I doubt it's a problem.

If you feel strongly about it, that line could be `${useNamespace(state)}sc-${getFileHash(state)}-${getNextId(state)}` to always insert `sc-` regardless of the namespace, but my feeling is that the namespace serves the purpose at that point, and the extra `sc-` is just noise.

chalbert:
Not being a maintainer, I'm not too sure of the impact. What I know is that jest-styled-components uses `sc-` to locate the class ― it was at the root of this bug as it was sometimes missing. So I guess it would be safer to keep `sc-` even if it does impose a bit of noise.

agriffis:
Sure, that makes sense. I just updated my diff above.

Main thing I'd like to avoid is extra complexity. There are two kinds of prefixing here: (1) for eyeballing and tools, (2) to avoid invalid class names. These different reasons were conflated by the original code. Rather than make that code more complex to cover more cases, it's better to tease them apart.

If you like it, please feel free to update your PR to use this approach.

chalbert:
@probablyup I simplified the solution.

akatakritos:
Thanks @chalbert -- I've been troubleshooting this all morning after an upgrade. Your forked package is working great for me now. 

Would love to get this merged in!

probablyup:
@chalbert  could you merge from main to get the tests running? Just enabled actions for forks. TY!

probablyup (approved):
Makes sense

Issue #322, opened 2021-04-28
Setting shouldForwardProp using withConfig breaks SSR componentId


ithinkdancan:
I noticed that when i try and use the `shouldforwardprop` method added in styled-components 5.1 it prevents this plugin from being able to add the componentId and displayName which is causing hydration errors when doing SSR. Is there way to merge the configs rather than just skipping over components using withConfig?

https://styled-components.com/docs/api#shouldforwardprop

PR #323, opened 2021-04-30
merge existing withConfig arguments to allow SSR when using shouldForwardProp

ithinkdancan:
Resolves issue #322 where if you have an existing withConfig on your component in order to leverage shouldForwardProp babel was not adding a displayName or componentId

this takes the existing expression adds the additional args and replaces the existing withConfig

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": true,
-  "trailingComma": "es5"
+  "trailingComma": "es5",
+  "arrowParens": "avoid"
 }
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -17,6 +17,7 @@ const addConfig = t => (path, displayName, componentId) => {
   }
 
   const withConfigProps = []
+
   if (displayName) {
     withConfigProps.push(
       t.objectProperty(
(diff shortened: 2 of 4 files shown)


1Seok2 (commented):
I think "optional chaining" improves the readability of the code. How about applying this?

ithinkdancan:
> I think "optional chaining" improves the readability of the code. How about applying this?

@1Seok2 I did not see optional chaining used anywhere else in the code so i instead matched convention with what was there. I assume that to support versions of node that do not support chaining.

ithinkdancan:
@probablyup can I get a review on this?

probablyup:
@ithinkdancan sorry for the delay in reviewing, can you also add a test case for `withConfig` and `attrs` being used together?

ithinkdancan:
@probablyup good call out, I have updated the code to be a little smarted about merging the withConfig arguments

probablyup (approved):
TY

probablyup:
@ithinkdancan could you merge from `main` to get the tests running? Just enabled actions for forks. TY!

ithinkdancan:
@probablyup rebased onto main and pushed it up.

probablyup:
> @probablyup rebased onto main and pushed it up.

Awesome, thank you.

probablyup:
@ithinkdancan waiting on a couple other small PRs and then will cut a release with this in it, thanks for your contribution!

Issue #280, opened 2020-04-16
Latest version not working with rollup ?


BerndWessels:
Hi

I put some `console.logs` into the `babel-plugin-styled-components` and it seems to get called but for some reason it does not process and the result is that for example `comments` and `whitespaces` are not removed from the `css templates` output, like in this resulting `minified output file`:

```js
"use strict";function e(){var n,r,t=(n=["\n  html {\n    // Antialiasing.\n    -webkit-font-smoothing: antialiased;\n    -moz-osx-font-smoothing: grayscale;  }\n"], ........
```

Can you see if I am doing something wrong or if there is a problem with `babel-plugin-styled-components` ? Thank you.

I have the following `rollup.config.js`

```js
import { terser } from "rollup-plugin-terser";
import autoExternal from "rollup-plugin-auto-external";
import babel from "rollup-plugin-babel";
import commonjs from "@rollup/plugin-commonjs";
import glob from "glob";

const components = glob.sync("./packages/*/src/*.js*");

export default components.map((component) => {
  return {
    input: component,
    output: {
      file: component.replace(/\/src\//g, "/lib/"),
      format: "cjs",
      plugins: [terser()],
    },
    plugins: [
      babel({
        exclude: "node_modules/**",
        plugins: ["babel-plugin-styled-components"],
        presets: ["@babel/preset-env", "@babel/preset-react"],
      }),
      commonjs(),
      autoExternal(),
    ],
  };
});
```

and these are the latest dependencies:

```json
{
  "author": "Design System Components Team",
  "browserslist": {
    "development": [
      "last 1 chrome version",
      "last 1 edge version",
      "last 1 explorer version",
      "last 1 firefox version",
      "last 1 safari version"
    ],
    "production": [
      "last 1 chrome version",
      "last 1 edge version",
      "last 1 explorer version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  },
  "dependencies": {
    "clsx": "^1.1.0",
    "prop-types": "^15.7.2",
    "react": "^16.13.1",
    "react-dom": "^16.13.1",
    "styled-components": "^5.1.0"
  },
  "description": "Design System Components",
  "devDependencies": {
    "@babel/core": "^7.9.0",
    "@babel/preset-env": "^7.9.5",
    "@babel/preset-react": "^7.9.4",
    "@rollup/plugin-commonjs": "^11.1.0",
    "@rollup/plugin-node-resolve": "^7.1.3",
[...]

PR #327, opened 2021-05-14
Resolve operability issue with rollup

ktranada:
When using `rollup` to bundle, all versions after `1.10.5` fail to generate a `withConfig` for styled components.

## Context
Reported issue [here](https://github.com/styled-components/babel-plugin-styled-components/issues/280).
Changes between versions: https://github.com/styled-components/babel-plugin-styled-components/compare/v1.10.5...v1.10.6

The breaking (but correct) change in `v1.10.6`:
```typescript
if (!bypassCache && cacheKey in localNameCache) {
  return localNameCache[cacheKey]
}
```

This change was necessary to utilize the cache and exit early:
```typescript
const localNameCache = {};

// First pass, cacheKey = 'first'
localNameCache[cacheKey]; // -> undefined
cacheKey in localNameCache; // -> false

// Second pass, cacheKey = 'first'
localNameCache[cacheKey]; // -> false
cacheKey in localNameCache; // -> true
```

Prior to `v1.10.6`,  the plugin would traverse the AST each time until a `localName` has been specified. Particularly, the code'll ignore previously stored `false` cache values.

## Investigation
As suggested in https://github.com/styled-components/babel-plugin-styled-components/issues/280#issuecomment-620301036, `rollup` will generate an AST from the provided source files. Not to go into too much detail, `rollup` appears to prepend some additional nodes to the AST.

I uploaded a gist that logs out each AST node that's considered by this plugin https://gist.github.com/ktranada/edad40fe3c01342eb6ededc3f2a4f669

suppose the following:
```typescript
// Box/index.tsx
import styled from 'styled-components'; // <- represented as an `MemberExpression ` node in the AST
export const Box = styled.div``;
```

When only using `babel` to transpile the code: 
- AST is generated
- Traverse the AST + parse nodes as described
  - The one and only `MemberExpression ` satisfies the requirements
  - the result is stored in the cache, using the filename as the cache key

When using `rollup` to bundle the code,
- Rollup generates its own AST of the source files
  - It may prepend nodes to the AST (these may or may not be present in the generated output)
- Traverse the AST
  - If rollup (or any bundler) prepends nodes to the AST and these nodes are parsed before the target `MemberExpression` node, the result is stored for the file and used for **all** subsequent lookups in this file -- this will cause the program to exit early with the stored result when parsing the target `MemberExpression` node.
  - In the gist above, you can see that there are multiple nodes to consider before reaching the node (line 155 of the gist) representing the `styled` import.

## Solution
The change introduced in `v1.10.6` is needed and should stay. 
We want to avoid repeatedly traversing the AST for known nodes and instead leverage the cached result.
We can augment the cache key with the name of the node that's being parsed.

### DIFF
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -14,8 +14,10 @@ export const isValidTopLevelImport = (x, state) =>
 
 const localNameCache = {}
 
-export const importLocalName = (name, state, bypassCache = false) => {
-  const cacheKey = name + state.file.opts.filename
+export const importLocalName = (name, state, options = {}) => {
+  const { cacheIdentifier, bypassCache = false } = options;
+  const cacheKeyAffix = cacheIdentifier ? `|${cacheIdentifier}` : '';
+  const cacheKey = name + state.file.opts.filename + cacheKeyAffix;
 
   if (!bypassCache && cacheKey in localNameCache) {
     return localNameCache[cacheKey]
@@ -74,9 +76,9 @@ export const isStyled = t => (tag, state) => {
   } else {
     return (
       (t.isMemberExpression(tag) &&
-        tag.object.name === importLocalName('default', state)) ||
+        tag.object.name === importLocalName('default', state, { cacheIdentifier: tag.object.name })) ||
       (t.isCallExpression(tag) &&
-        tag.callee.name === importLocalName('default', state)) ||
+        tag.callee.name === importLocalName('default', state, { cacheIdentifier: tag.callee.name })) ||
       /**
        * #93 Support require()
        * styled-components might be imported using a require()
--- a/src/visitors/transpileCssProp.js
+++ b/src/visitors/transpileCssProp.js
@@ -44,7 +44,7 @@ export default t => (path, state) => {
       nameHint: 'styled',
     })
 
-    importName = t.identifier(importLocalName('default', state, true))
+    importName = t.identifier(importLocalName('default', state, { bypassCache: true }))
   }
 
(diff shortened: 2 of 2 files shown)


ktranada:
I didn't add tests because this requires running rollup to comparing results, but the testing architecture only captures babel transpilations
I'd be more than happy to add tests if there are any suggestions to introduce it into the repo :)

src/utils/detectors.js
```diff
@@ -14,8 +14,10 @@ export const isValidTopLevelImport = (x, state) =>
 
 const localNameCache = {}
 
-export const importLocalName = (name, state, bypassCache = false) => {
-  const cacheKey = name + state.file.opts.filename
+export const importLocalName = (name, state, options = {}) => {
+  const { cacheIdentifier, bypassCache = false } = options;
+  const cacheKeyAffix = cacheIdentifier ? `|${cacheIdentifier}` : '';
+  const cacheKey = name + state.file.opts.filename + cacheKeyAffix;
```
ktranada, line 20: I'm not married to the idea of affixing the value, so long as it's included in the `cacheKey`

iDVB:
Any luck with this? We've [run into this recently](https://github.com/styled-components/babel-plugin-styled-components/issues/280#issuecomment-827177595) and fixed with the downgrade to `babel-plugin-styled-components@v1.10.5` for most of our project. However, we still have one project that is already downgraded to `v1.10.5` and still having this same issue.

probablyup (approved):
LGTM

PR #338, opened 2021-07-05
Safe check for argument properties

ghoshArnab:
Screenshot 2021-07-05 at 6 10 51 PM

### DIFF
--- a/src/visitors/displayNameAndId.js
+++ b/src/visitors/displayNameAndId.js
@@ -40,7 +40,7 @@ const addConfig = t => (path, displayName, componentId) => {
   if (
     existingConfig &&
     existingConfig.arguments.length &&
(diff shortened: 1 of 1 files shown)


probablyup:
@ghoshArnab do you happen to have a code sample of where it broke so I can add a regression test?

probablyup:
@ghoshArnab just checked and this isn't available until Node 14, so we can't use this syntax for the time being because the 1.x semver branch doesn't have a node "engines" requirement. Could you switch the code to something more compatible?

ghoshArnab (commented):
Removed optional chaining operator

probablyup:
Added these in 0c16c1bac6b2a86bac66cf2527f1b05b896019b8, thanks!

PR #340, opened 2021-07-06
handle namespace import helpers

### LINKED ISSUE #315: (...).createGlobalStyle.withConfig is not a function
Similar to this [comment workaround](https://github.com/styled-components/babel-plugin-styled-components/issues/216#issuecomment-516941240), when i use createGlobalStyle like this:

`import * as styled from 'styled-components';`
`export const GlobalStyle = styled.createGlobalStyle``;`, i get the error `(...).createGlobalStyle.withConfig is not a function` on a Next.js / Typescript project

The bug is resolved when I use the solution presented in the comment above, using CommonJS require:
`const createGlobalStyle = require('styled-components').createGlobalStyle`
`export const GlobalStyle = createGlobalStyle``;`

Is it a bug or a setup issue?

probablyup:
Fixes #315 

Handle the following syntax:

```js
import * as styled from 'styled-components'

const css = styled.css`
  background: black;
`

const GlobalStyle = styled.createGlobalStyle`
  html {
    background: black;
  }
`

const Test = styled.default.div`
  color: red;
`

const before = styled.default.div`
  color: blue;
`

styled.default.div``

export default styled.default.button``

```

Previously `styled.default.*` was not being assigned identifiers, and helpers were being assigned identifiers when they should not be.


RobbyUitbeijerse:
@probablyup coincidentally, this update also fixed our situation where we use a cjs Rollup build (external design system) with both babel-plugin-styled-components as well as next-transpile-modules. It resulted in [this error](https://github.com/styled-components/styled-components/issues/2681), but is resolved after we installed the test release from #340 :-)

Thanks for the update!

PR #354, opened 2021-09-13
Change topLevelPaths option to take a regexp pattern

rockwotj:
This allows for local imports like ../../resources/styled where the relative path changes so the existing pattern doesn't work.


probablyup (approved):

agriffis:
This change resulted in some fallout, see https://github.com/styled-components/styled-components/issues/3635

Regexes are more powerful, but more confusing and prone to error. Either of the other suggestions (picomatch or typed objs) would have been a better approach, and would have avoided the breaking change.

As an example of prone to error, this PR uses `re.test` which isn't anchored, so now strings in `topLevelImportPaths` get converted to patterns that match anywhere in the input. The modified test even demonstrates the potential for mistakes:

           "@xstyled/styled-components/*"

This pattern uses `*` instead of `.*` but it matches `'@xstyled/styled-components/test'` anyway because `re.test` isn't anchored.

@rockwotj @probablyup It seems clear that pinning `styled-components` to version 1 of `babel-plugin-styled-components` isn't the best fix, even if it's a temporary workaround for the linked issue. Do you have thoughts on the best path forward?

PR #359, opened 2021-11-21
handle complex expressions inside CSS prop object syntax

### LINKED ISSUE #352: Usage of css prop broken in 1.13.2
Hi,

Upgrading this library from v1.12.0 to v1.13.2 is breaking my app builds. I've pinned it down to a specific component that it's using the `css` prop as follows:
```
const MyComponent = ({ width, ...props }) => {
  return (
    <ThirdPartyComponent css={{ width: width ? `${width}px` : '100%' }} {...props} />;
  );
};
```
For the record, the issue still happens when using an HTML tag (i.e. `div`) instead of a custom `ThirdPartyComponent`.

Is this a bug or is it intended behaviour after the upgrade?

**Note:** I'm building my application with the latest versions of Webpack 5 and Babel 7, and using styled-components@5.2.3

probablyup:
fixes #352

### DIFF
--- a/src/visitors/transpileCssProp.js
+++ b/src/visitors/transpileCssProp.js
@@ -224,58 +224,18 @@ export default t => (path, state) => {
       ) {
         replaceObjectWithPropFunction = true
 
-        if (
-          // handle tricky expressions
-          t.isConditionalExpression(property.value) ||
-          t.isLogicalExpression(property.value)
-        ) {
-          const jSXIdentifier = getLocalIdentifier(path)
-          const injectedIdentifier = getLocalIdentifier(path)
+        const identifier = getLocalIdentifier(path)
 
-          let insertionPoint = elem
-
-          while (insertionPoint.key !== 'body') {
-            insertionPoint = insertionPoint.parentPath
-          }
-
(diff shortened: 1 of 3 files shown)


klarstrup:
One of these changes broke my thing that isn't using the object syntax, I hope to find out more later but just an immediate FYI

probablyup:
@klarstrup must be something to do with this change: https://github.com/styled-components/babel-plugin-styled-components/pull/359/files#diff-ef6a6ffba3d725918f184d76241892f9e7ac878aa3891cce860feea7648794b3L125

If you can come up with an example of the syntax that triggers the error, would appreciate it.

klarstrup:
You'll have to excuse the cobbled together repro but here you see the problem at a glance:

v2.0.1
image

v2.0.0
image

klarstrup:
So in short there's an issue of spread `...props` disappearing when combined with `css` within a render prop (but this also happens if one copies out the render prop function into for example a `useMemo`)

klarstrup:
fix is in btw https://github.com/styled-components/babel-plugin-styled-components/pull/361

PR #364, opened 2021-11-29
Add failing test case

Janpot:
Ran into this while debugging another problem. `babel-plugin-styled-components` doesn't recognise these imports:
```tsx
import { styled as matStyled } from '@mui/material/styles'
```

### DIFF
--- a/test/fixtures/mui/.babelrc
+++ b/test/fixtures/mui/.babelrc
@@ -1,0 +1,19 @@
+{
+  "plugins": [
+    [
+      "../../../src",
+      {
+        "fileName": false,
+        "transpileTemplateLiterals": false,
+        "ssr": true,
+        "topLevelImportPaths": [
+          "@mui/material",
+          "@mui/material/*",
+          "@mui/system",
+          "@mui/styled-engine-sc"
+        ]
+      }
+    ],
+    ["@babel/plugin-proposal-class-properties", { "loose": true }]
+  ]
+}
--- a/test/fixtures/mui/output.js
+++ b/test/fixtures/mui/output.js
@@ -1,0 +1,14 @@
+import { styled as matStyled } from '@mui/material/styles';
+import sc from '@mui/material/styles';
+const Test1 = matStyled('div').withConfig({
+  displayName: 'Test1',
+  componentId: 'sc-2jen0y-0',
+})({
+  width: '100%'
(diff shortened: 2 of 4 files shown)


test/fixtures/mui/.babelrc
```diff
@@ -0,0 +1,19 @@
...
+  "plugins": [
+    [
+      "../../../src",
+      {
+        "fileName": false,
+        "transpileTemplateLiterals": false,
+        "ssr": true,
+        "topLevelImportPaths": [
+          "@mui/material",
+          "@mui/material/*",
```
rockwotj, line 11: FWIW at HEAD right now this doesn't work. You need a regex. I have a PR to change this to picomatch

MonstraG:
Not OP, but this PR is mentioned in open https://github.com/mui/material-ui/issues/29742, so I guess it is.

test/fixtures/mui/output.js
```diff
@@ -0,0 +1,14 @@
+import { styled as matStyled } from '@mui/material/styles';
+import sc from '@mui/material/styles';
+const Test1 = matStyled('div').withConfig({
```
rbong, line 3: Do you really want it to transform `matStyled`? According to this issue this isn't wanted: https://github.com/styled-components/babel-plugin-styled-components/issues/224

mnajdova: We didn't want to do this in v4, because the `styled` there was not using `styled-components`. For v5, when the `@mui/system` is configured to use `styled-components`, we want the plugin to process these imports.

slavco86:
@Janpot , @rbong  - any update on this, pleae? 🙏

landsman:
Hello, is there a plan how to fix this?

quantizor:
Thanks for setting this up, @Janpot. To recognize a styled import that doesn't come from `styled-components` itself (MUI's `@mui/material/styles` in your case), the plugin needs the import path on the recognized list. Today the supported way to do that is the `topLevelImportPaths` option:

  ['babel-plugin-styled-components', { topLevelImportPaths: ['@mui/material/styles'] }]

With that, MUI's `styled` export gets the same treatment as the default `styled-components` import. Closing this in favor of that route. The fixture stays useful as a reference if first-class MUI integration ever lands.

PR #365, opened 2021-11-30
Doesn't work on the MUI commonjs modules

Janpot:
I'm currently investigating an issue on `@mui/material`+`styled-components`+`next` (https://github.com/mui-org/material-ui/issues/29742). It turns that in the Next.js server bundle, the `styled` calls are not correctly being identified by `babel-plugin-styled-components`. The server bundle uses the commonjs version of `@mui/material`. This PR adds a failing test case for that format.

_edit: 🤔  The new test doesn't seem to run in CI for some reason_

### DIFF
--- a/test/fixtures/compiled/.babelrc
+++ b/test/fixtures/compiled/.babelrc
@@ -1,0 +1,13 @@
+{
+  "plugins": [
+    [
+      "../../../src",
+      {
+        "fileName": false,
+        "transpileTemplateLiterals": false,
+        "ssr": true
+      }
+    ],
+    ["@babel/plugin-proposal-class-properties", { "loose": true }]
+  ]
+}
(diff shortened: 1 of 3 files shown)


slavco86:
@Janpot , any update on this?

landsman:
Hello, is there a plan how to fix this?

quantizor:
Thanks @Janpot. Same situation as #364: MUI's `@mui/material/styles` is not on the plugin's default import-path list, so the `styled` export from there isn't recognized without opt-in. The `topLevelImportPaths` option is how to wire it up:

  ['babel-plugin-styled-components', { topLevelImportPaths: ['@mui/material/styles'] }]

Closing in favor of that route. The CJS-shape fixture you put together is a useful regression reference; thank you for the investigation.

Issue #350, opened 2021-08-22
How use with Vite.js, especialy displayNmae for debugging.


valerii15298:
When using with CRA, its possible to see name of component inside css classes, but I cannot figure out how to make a nice developing/debugging experience when using with Vite.js. Is there any advices how to show in classes names of components when using with Vite.js?

rockwotj:
I suggest using https://www.npmjs.com/package/vite-plugin-babel-macros and importing from [`styled-components/macro`](https://styled-components.com/docs/tooling#babel-macro). We have this working nicely in our vite setup (although that's with a custom plugin that should do the same as the package I pointed you too).

ronaldruzicka:
Hi @rockwotj, please could you share your code how you set it up?

I have this config
```js
// vite.config.ts
import { defineConfig } from 'vite';
import reactRefresh from '@vitejs/plugin-react-refresh';
import macrosPlugin from "vite-plugin-babel-macros"

export default defineConfig({
  plugins: [
    reactRefresh(),
    macrosPlugin(),
  ],
});
```
And then I just import from `styled-components/macros`
```js
// App.tsx

import styled from 'styled-components/macros';
```

But I get a Typescript error 
```

**Comment by rockwotj** [comment]
> Hi @rockwotj, please could you share your code how you set it up? 
> Or could you please tell me what am I doing wrong?
> 

Looks like it's not installed. Did you run npm install?

**Comment by ronaldruzicka** [comment]
Oh my bad, I was using plural `/macros` instead of singular `/macro` 🤦‍♂️ 
Now it's working properly, thanks a lot!

**Comment by IvanBernatovic** [comment]
@ronaldruzicka I added the macros plugin and started importing styled like this:
`import styled from 'styled-components/macro`

But I still don't get proper class names, they are still randomized/minified. Is there anything else I should do?

**Comment by rockwotj** [comment]
@IvanBernatovic you need to have a config file to get display names, see the docs for more: https://styled-components.com/docs/tooling#experimental-config

**Comment by ronaldruzicka** [comment]
@IvanBernatovic I didn't use any other config. I just included that `macrosPlugin` in the `vite.config.ts` and then used `import styled from 'styled-components/macro';` in the component and it worked.

I have these versions:
```
"styled-components": "^5.3.0"
"vite-plugin-babel-macros": "^1.0.5",
```

But maybe try that experimental config as well.

IvanBernatovic:
[...]

PR #369, opened 2021-12-29
Use picomatch instead of regex

rockwotj:


test/fixtures/add-identifier-with-top-level-import-paths/code.js
```diff
@@ -1,13 +1,12 @@
-import styledA from '@xstyled/styled-components/test'
-import styledB from '@xstyled/styled-components-test'
-import styledC from '@xstyled/styled-components'
+import styled from '@xstyled/styled-components'
+import unstyled from '@xstyled/styled-components-test'
```
agriffis, line 3: I think you might want these three:
- `@xstyled/styled-components` (direct match)
- `@xstyled/styled-components/test` (globbed match)
- `@xstyled/styled-components-test` (doesn't match)

Those were the originals, I guess. What was the motivation to change it?

agriffis:
Apart from my minor question about the test above, this looks great to me.

agriffis:
As mentioned in #369, this will need to bump the major version of the plugin. Additionally we might want to consider changing the dep string in `styled-components` for the macro to `^1 || ^3` or along those lines.

test/fixtures/add-identifier-with-top-level-import-paths/code.js
rockwotj: This plugin only checks the first import match than bails as far as I can tell. It only matches one and reordering the import causes a different one to be selected. I've decided to limit this test to the simple strict match and the other test can be for testing the picomatch behavior

agriffis: Sounds good!

agriffis (approved):

agriffis:
Rounding up some of the issues that will be fixed by this:

https://github.com/styled-components/styled-components/issues/3635
https://github.com/styled-components/styled-components/issues/3645
https://github.com/styled-components/babel-plugin-styled-components/issues/350

agriffis:
Just need to get @probablyup to notice it :eyes:

probablyup (approved):
Thanks!

davidcostadev:
OMG, this fixed my issue, thanks! I'm using it for displayNames on the react-scripts env <3

Issue #366, opened 2021-12-13
minifier removes space between & and ::after


agriffis:
There's a bug reported against SWC because it works differently from this plugin, but AFAICT the bug is in the plugin, not SWC.

Report: https://github.com/vercel/next.js/issues/30802#issuecomment-990960985
My response: https://github.com/vercel/next.js/issues/30802#issuecomment-992766906

Basically when the minifier encounters `.globalClass & ::after` it kills the space between the ampersand and the colons.

jukkah:
I have the same problem with `& :not(:last-child)`. Also tested with some other selectors.

Context: In my case (https://github.com/vrk-kpa/suomifi-ui-components/blob/v7.0.0-beta.1/src/core/Form/FilterInput/FilterInput.baseStyles.tsx#L39) I'm trying to add separator between items. But after minification the selector matches to the parent element instead of the items.

agriffis:
Fixed by #376 and released in v2.0.7

PR #376, opened 2022-04-05
fix minifier bug #366

agriffis:
The minifier doesn't distinguish between selectors and rules, so we preserve any whitespace separator preceding a colon.


agriffis:
@probablyup :eyes:

probablyup (approved):
LGTM thank you

Issue #379, opened 2022-07-30
CommonJS imports not identified


rbong:
Imports which use `_interopRequireDefault` are not identified.

rbong:
This is what causes https://github.com/styled-components/babel-plugin-styled-components/pull/365

eckmLJE:
We have run into this issue using module federation and our in-house component library. babel-plugin-styled-components successfully namespaces styled components written within each app, preventing class name collision when multiple apps are loaded on the same page with module federation. However, each app also imports from our component library which is already transpiled, and these component class names are not namespaced, leading to class name collision bugs. 

Commenting to add our support for @rbong's [PR](https://github.com/styled-components/babel-plugin-styled-components/pull/380) to add this support after testing it in our dev environment.

quantizor:
Hi @rbong, thanks for filing this and apologies for the delay.

Both shapes you described are handled on current main. The existing `pre-transpiled` fixture covers `_interopRequireDefault(require('styled-components'))`, and the bare `(0, styled)('div')` SequenceExpression form is handled by the same detection arms in `src/utils/detectors.js` (verified locally with a fresh fixture).

Closing as resolved.

PR #380, opened 2022-07-30
Support pre-transpiled code (#379)

rbong:
This adds support for identifying styled components imports using `_interopRequiredDefault` and calls to styled components that use sequence expressions (ex. `(0, _styled)('div')()`).

This should improve support for MUI + Next with SSR.

In `isStyled`, I added equivalent checks for sequence expressions wherever they could be used instead of member expressions. However, there's one case that I'm not sure will ever apply:

```js
(importLocalName('default', state) &&
  t.isCallExpression(tag) &&
  t.isMemberExpression(tag.callee) &&
  tag.object.property.name === 'default' &&
  tag.object.object.name === importLocalName('default', state))
```
  
Please let me know if this needs an equivalent sequence expression check.
  
I thought maybe this PR might add some new errors on re-transpilation, but I ran tests using outputs as inputs and didn't see any new problems.

### DIFF
--- a/src/utils/detectors.js
+++ b/src/utils/detectors.js
@@ -1,3 +1,4 @@
+import getSequenceExpressionValue from './getSequenceExpressionValue'
 import { useTopLevelImportPathMatchers } from './options'
 
 const VALID_TOP_LEVEL_IMPORT_PATH_MATCHERS = [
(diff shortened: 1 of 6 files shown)


eckmLJE:
Re-posting this from the [issue](https://github.com/styled-components/babel-plugin-styled-components/issues/379). 
>We have run into this issue using module federation and our in-house component library. babel-plugin-styled-components successfully namespaces styled components written within each app, preventing class name collision when multiple apps are loaded on the same page with module federation. However, each app also imports from our component library which is already transpiled, and these component class names are not namespaced, leading to class name collision bugs. 

We've tested this PR in our dev environment and it fixes our issue.

probablyup (approved):
Omg thank you, sorry I didn’t see this sooner

PR #381, opened 2022-10-14
Add tricky expression on object expression

healtheloper:
except object key interpolation if property value is member expression

```tsx
const ObjectInterpolationMember = p => {
  const theme = useTheme()
  const color = 'red'

  return (
    <p
      css={{
        color: theme.colors[color],
      }}
    >
      H
    </p>
  )
}
```

스크린샷 2023-04-06 오전 2 44 34


healtheloper:
@probablyup 👀

probablyup (approved):
Thanks!

PR #389, opened 2023-05-10
When testing import paths, only use string literals.

maisano-patreon:
Prior to this change, code with template literals inside of `require` calls would break. This is because the check of `t.isLiteral` is far too broad for this usage, and assumes that all literals have a `value` property. Since `TemplateLiteral` nodes do not, they would cause `picomatch` to fail on an `undefined` input.


maisano-patreon:
@probablyup any thoughts on this?

probablyup (approved):
LGTM!

PR #410, opened 2024-09-14
Replace lodash with es-toolkit

wojtekmaj:
es-toolkit is a modern lodash alternative that claims to be much faster and lighter, potentially improving overall install and runtime performance: https://es-toolkit.slash.page/intro.html


quantizor:
Thank you for opening this @wojtekmaj. The nudge toward dropping the lodash dependency was good. I ended up removing lodash outright in 2.2.0 rather than swapping to es-toolkit, so this swap is no longer needed. Closing as no-longer-applicable, with thanks.

